PR #2, opened 2026-05-16
Misc fixes

amadeus:
Fix a few issues with codiff.

* Updated to use the latest beta which fixed a layout calculation bug when using `simple` hunk separators
  * Also includes a bunch of other perf improvements and fixes
 * Reworked padding to properly align with new box-shadow borders
* The custom headers don't use unsafeCSS, and instead use global css because of slots api, so removed those unsafeCSS declarations
  * Fixed up custom headers to have both a cleaner selected state and be look better when collapsed
* Fixed how scrollTo worked both with new paddings and with automatic scroll handling
  * Before scrolling to a file only worked after you manually scrolled
  * Also if you clicked on an item normal scroll would trigger effects that would scroll you back to the selected file
  * anyways fixed all that (mostly vibe coded tho, but it does work)
* Polished interactions with collapsed vs viewed
  * Marking as viewed always collapses refreshing app will maintain that viewed state)
  * Collapsing acts independently of viewed state, and won't adjust it and you can still expand once viewed
* Reworked versioning system to fit with beta.5
  * It's just a number now, that should increment
* Fixed scrollable region to be full height with no inner padding cropping of the code content
* Fixed default scrollbars in dark theme (they now properly theme with your OS)

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "test": "vp test"
   },
   "dependencies": {
-    "@pierre/diffs": "1.2.0-beta.5",
+    "@pierre/diffs": "1.2.0-beta.6",
     "@pierre/trees": "1.0.0-beta.3",
     "electron": "^42.0.1",
     "react": "^19.2.6",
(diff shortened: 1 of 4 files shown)


cpojer:
Amazing, thank you!

PR #3, opened 2026-05-16
Add project switcher

FllipEis:
## Description
Codiff currently opens a single repository per window and requires launching the app again to inspect a different project. That works for the smallest workflow, but it makes switching between nearby repositories awkward. This PR adds a lightweight project switcher so users can open another repository from the sidebar, switch back to recently opened repositories, and have the last selected project restored on launch.

This is intentionally small, but I am not fully sure yet what structure fits the project best as it grows. I split the project state handling into a small Electron-side project manager and the sidebar control into a React component, but I would be happy to adjust naming or structure if there is a preferred convention for this codebase.

## Changes
- Added Electron IPC handlers for reading known projects, selecting a project, and opening a repository directory through the native folder picker.
- Added `electron/project-manager.cjs` to track the active repository per window, normalize project roots, validate repositories, persist recent projects, and restore the last project.
- Extended preferences with `lastProjectRoot` and `projects`.
- Updated repository state, history, show-in-folder, and relative-path IPC handlers to use the active project for the current window.
- Added a `ProjectSelector` React component for the sidebar repository dropdown and open-repository button.
- Refactored repository loading in `App.tsx` so project switching resets repository-specific UI state and loads viewed/collapsed state for the selected repository.
- Added `ProjectEntry` and `ProjectList` types plus preload/global API definitions for the new project APIs.
- Updated sidebar styles for the project selector and open button.

## Type of Change
- [x] New feature / capability
- [ ] Bug fix
- [ ] Refactoring
- [ ] Documentation
- [ ] Infrastructure / CI/CD

## Behavior Change
Codiff can now switch repositories from within the app instead of being tied to only the repository selected at launch.

**Before:**
The sidebar showed the current repository path as static text. To inspect another repository, the user had to relaunch/open another Codiff instance with a different path.

**After:**
The sidebar shows a repository selector. Users can open another repository with the native folder picker, switch between known repositories from the dropdown, and Codiff restores the last selected repository on the next launch unless an explicit launch path is provided.

## Pre-Deployment Migrations
No migrations required before deploying.

### DIFF
--- a/electron/main.cjs
+++ b/electron/main.cjs
@@ -28,11 +28,13 @@ const {
   readRepositoryChangeSignature,
   readRepositoryState,
 } = require('./git-state.cjs');
+const { createProjectManager } = require('./project-manager.cjs');
 
 const root = dirname(__dirname);
 const repositoryWatchers = new Map();
-const windowRepositories = new Map();
 let preferences = {
+  lastProjectRoot: null,
+  projects: [],
   showWhitespace: false,
 };
 
@@ -41,8 +43,11 @@ const getCommandLineRepositoryPath = (commandLine = process.argv) => {
   return args.find((arg) => arg && !arg.startsWith('-'));
 };
 
+const getExplicitLaunchPath = () =>
+  process.env.CODIFF_REPOSITORY_PATH || getCommandLineRepositoryPath();
+
 const getLaunchPath = () =>
-  resolve(process.env.CODIFF_REPOSITORY_PATH || getCommandLineRepositoryPath() || process.cwd());
+  resolve(getExplicitLaunchPath() || preferences.lastProjectRoot || process.cwd());
 
 const getPreferencesPath = () => join(app.getPath('userData'), 'preferences.json');
 
@@ -60,6 +65,17 @@ const readPreferences = () => {
 const writePreferences = () => {
   writeFileSync(getPreferencesPath(), JSON.stringify(preferences, null, 2));
 };
+
+const updatePreferences = (updater) => {
+  preferences = updater(preferences);
+  writePreferences();
+};
+
+const projectManager = createProjectManager({
+  getLaunchPath,
+  getPreferences: () => preferences,
+  updatePreferences,
+});
 
 const sendPreferencesChanged = () => {
   for (const window of BrowserWindow.getAllWindows()) {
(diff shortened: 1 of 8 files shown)


cpojer:
Hey! Have been working on it actively, so there are merge conflicts now and I haven't checked out your branch yet. I just added a `File → Open Folder…` feature on main, so not sure we need this switcher still? What do you think?

FllipEis:
> Hey! Have been working on it actively, so there are merge conflicts now and I haven't checked out your branch yet. I just added a `File → Open Folder…` feature on main, so not sure we need this switcher still? What do you think?

Hey!

I saw the new `File → Open Folder…` flow, and it definitely overlaps with part of this PR. I still think there’s value in switching projects within the same window though. The sidebar switcher is not strictly necessary, but it could make the workflow nicer by avoiding a new window for every repository, especially when working across multiple projects.

If the sidebar does not feel like the right design or position for this, another option could be to ask whether a folder should open in the current window or a new window when using `File → Open Folder…`. We could also add a recent folders/projects entry somewhere, so switching back to an existing project is still easy.

If the direction is to keep one project per window, that’s totally fine too. Happy to close this or adjust it however you think makes the most sense. 😄

cpojer:
Yep, I want one window per repo at this point in time. I am just vibing this, so feel free to fork it and make it work for your use cases specifically. Really appreciate your contribution, feel free to reach out on Twitter or elsewhere to discuss future direction if you want to keep contributing :)

FllipEis:
> Yep, I want one window per repo at this point in time. I am just vibing this, so feel free to fork it and make it work for your use cases specifically. Really appreciate your contribution, feel free to reach out on Twitter or elsewhere to discuss future direction if you want to keep contributing :)

Totally understand, thanks for clarifying the direction.

I still really like this project and would love to help where I can. I also DM’d you on Twitter yesterday and would be happy to chat about future direction or contributions there. Thanks again!

electron/main.cjs
```diff
@@ -346,26 +371,59 @@ if (squirrelStartup || !lock) {
...
 
 ipcMain.handle('codiff:getRepositoryHistory', async (event, limit) => {
-  const repositoryPath = windowRepositories.get(event.sender.id) || getLaunchPath();
+  const repositoryPath = projectManager.getActiveRepositoryPath(event.sender.id);
   return listRepositoryHistory(repositoryPath, limit);
 });
 
+ipcMain.handle('codiff:getProjects', async (event) =>
+  projectManager.readProjectList(event.sender.id),
+);
```
cpojer, line 392: Test Comment (Please ignore).

PR #6, opened 2026-05-18
feat: export review comments for automation

winter-loo:
## Summary
- export in-memory review comments to `.codiff/review-comments.md` and `.codiff/review-comments.json`
- expose a small Electron IPC bridge so external review automation can consume reviewer feedback
- keep native window controls visible on Windows/Linux while preserving macOS hidden title bar behavior

## Test plan
- `npx --yes pnpm@11.0.6 test`
- `npx --yes pnpm@11.0.6 exec vp build`

### DIFF
--- a/electron/main.cjs
+++ b/electron/main.cjs
@@ -347,8 +347,12 @@ const createWindow = (repositoryPath, launchOptions = { walkthrough: false }) =>
     minWidth: 880,
     show: false,
     title: `Codiff - ${repositoryPath}`,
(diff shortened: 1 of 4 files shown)


cpojer:
Thanks for the PR! I adopted the title bar style changes. However, I don't want to automatically write the review comments. Right now you can copy them into your agent with one click, and the new version allows you to ask an agent directly inline. I think we should just make more direct integrations to talk back and forth with an agent.

PR #8, opened 2026-05-19
Allow Codex model override via env

tobiasandersen:
I couldn't use Walkthrough with my subscription because it rejected the hardcoded `gpt-5.3-codex-spark` model, so this changes Codiff to let Codex use its local default unless `CODIFF_CODEX_MODEL` is set explicitly.

**And thanks for building this!**

### DIFF
--- a/electron/codex.cjs
+++ b/electron/codex.cjs
@@ -4,7 +4,6 @@ const { tmpdir } = require('node:os');
 const { join } = require('node:path');
 
 const CODEX_TIMEOUT_MS = 45_000;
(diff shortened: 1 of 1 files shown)


cpojer:
Love the idea! Sorry didn't realize spark wasn't available in some plans.

I didn't love the idea of using env vars, so the next version will detect the available models + provide a picker:

CleanShot 2026-05-19 at 17 13 02@2x

(fyi 5.3 spark appeared to do the best & fastest job for code reviews, which is why an older model is the default. It's intentional)

PR #11, opened 2026-05-19
Add option to automatically copy comments to clipboard on quit

bryceosterhaus:
Another random bit that I find myself using so that I don't have to click the copy button. Again, feel free to disregard if you don't think it is helpful or aligns with the app.

By default it is not enabled, but here is a shot of it enabled.

Screenshot 2026-05-19 at 1 25 16 PM

----

**Demo:** (I used CMD+w to close codiff)

https://github.com/user-attachments/assets/47b8d3ff-1a3a-4a4a-8f36-331436b4a771

### DIFF
--- a/electron/main.cjs
+++ b/electron/main.cjs
@@ -17,6 +17,7 @@ const { parseArgs } = require('node:util');
 const {
   app,
   BrowserWindow,
+  clipboard,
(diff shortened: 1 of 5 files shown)


cpojer:
Cool feature.

PR #14, opened 2026-05-20
A few more UI improvements

bryceosterhaus:
## Summary
This PR adds a configurable keymap, a command bar, a collapsible sidebar, and several UI polish items (history avatars/dates, copy-file-path button, copy-and-close command).
### Central config file with configurable keymaps
Introduces `~/.codiff/codiff.jsonc` as the source of truth for user settings, including a fully configurable keymap. Ships with a JSON schema for editor autocomplete and validation, hot-reloads on file change, and migrates existing preferences automatically.
### Command bar (Cmd+Shift+P)
Adds a global command registry and a VS Code-style command bar. Supports fuzzy filtering, arrow-key navigation, and displays the bound keyboard shortcut next to each command. Commands include file-filter focus, find-in-diffs, sidebar mode switching, and sidebar toggle.
### Collapsible sidebar
The sidebar can now be collapsed via a toggle button or `Cmd+B`. The collapsed state persists across launches, and the top bar animates smoothly. Activating `Cmd+P` while collapsed auto-expands and focuses the file filter.
### History avatars and metadata
History entries now show the author's Gravatar, name, and commit date. The repository's current branch is displayed in the sidebar header.
### Copy file path button
Adds a small button next to each file path in the diff view to copy the path to clipboard.

-------
**Demo:**

https://github.com/user-attachments/assets/11c33a8d-3a1f-4946-9159-18559e74d117

### DIFF
--- a/electron/config.cjs
+++ b/electron/config.cjs
@@ -33,6 +33,7 @@ const defaultSettings = {
 /** @type {CodiffKeymap} */
 const defaultKeymap = {
   closeSearch: 'Escape',
+  commandBar: 'Mod+Shift+p',
   diffSearch: 'Mod+f',
   discardComment: 'Escape',
   fileFilter: 'Mod+p',
@@ -152,6 +153,8 @@ const mergeConfig = (raw) => {
         typeof rawKeymap.closeSearch === 'string'
           ? rawKeymap.closeSearch
           : defaultKeymap.closeSearch,
+      commandBar:
+        typeof rawKeymap.commandBar === 'string' ? rawKeymap.commandBar : defaultKeymap.commandBar,
       diffSearch:
         typeof rawKeymap.diffSearch === 'string' ? rawKeymap.diffSearch : defaultKeymap.diffSearch,
       discardComment:
--- a/electron/config.cjs
+++ b/electron/config.cjs
@@ -40,6 +40,7 @@ const defaultKeymap = {
   nextSearchMatch: 'Enter',
   prevSearchMatch: 'Shift+Enter',
   submitComment: 'Mod+Enter',
+  toggleSidebar: 'Mod+b',
 };
 
 /** @type {CodiffConfig} */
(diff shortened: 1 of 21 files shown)


cpojer:
Feedback:
* Add a gravatar component with a `size: medium | small option`, use it in the history view and comments.
* I'm not a fan of the header change. While Codiff certainly doesn't look like a Liquid Glass native app, this header makes it look more outdated than it is now. Are you married to this idea or ok to revert for now?
* The copy button needs to go a bit further up (I think we also need to figure out icons since the transparency looks terrible on icons right now).

Feel free to bring the changes from https://github.com/nkzw-tech/codiff/issues/15 into this PR as well, it's easier to just have one PR that I can work with.

bryceosterhaus:
> Are you married to this idea or ok to revert for now?

 Fine to revert it, I had it locally because when collapsing the sidebar, the traffic lights on macos need a place to live and the diffed file accordion overlapped with it.
 
 > The copy button needs to go a bit further up (I think we also need to figure out icons since the transparency looks terrible on icons right now).
 
 👍

cpojer:
I think for collapsing the sidebar it would be awesome if it animated the sidebar down to the essential of traffic lights + title + expand button, and at the same time pushes the top of the app down so the sidebar does not overlap with the content and effectively becomes a top bar. Not sure if that's possible smoothly, but would be sick.

cpojer:
@bryceosterhaus are you doing a pass on this PR still or not?

bryceosterhaus:
Yeah still adding the config and command bar. I will update it here in the next hour or two

bryceosterhaus:
Updated the menu bar on collapse and also added the things discussed in https://github.com/nkzw-tech/codiff/issues/15.

Let me know what you think

bryceosterhaus:
The icons still looks off and can be misaligned, probably just need to use an icon set for consistency in another PR rather than having llm's generate some ad-hoc icons

cpojer:
Awesome! Thank you. Just a question: which llm do you use to write code?

bryceosterhaus:
Mostly opus 4.6 and 4.7 right now via opencode, sometimes jumping to gpt5.5 if opus is giving me strange output

cpojer:
Got it! Highly recommend using gpt 5.5 for Codiff PRs going forward.

cpojer:
Merged the PR with various improvements. This was a great contribution again, thank you so much for making the app better every day!

PR #16, opened 2026-05-20
Fix Cmd+F search highlights after diff library DOM change.

bryceosterhaus:
Wasn't working on main

**After:**
Screenshot 2026-05-20 at 2 53 27 PM


cpojer (approved):
Nice!

PR #17, opened 2026-05-20
fix: parse hash-prefixed PR marker values

lyzno1:
## Summary
- Accept hash-prefixed values in `codiff pr #123`.
- Keep packaged CLI parsing aligned with Electron command-line parsing.


cpojer (approved):
Makes sense, thank you!

PR #18, opened 2026-05-20
Resolve Codex through the user's shell

LarsEckart:
Why it crashed for me: the `codex` on my path still pointed to a global npm install that didn't exist anymore and thus it failed inside Codiff.

I've installed codex through npm which is installed through [n](https://github.com/tj/n) which is installed through brew.

I got it working with Pi and GPT-5.5, but it felt like it changed a bit too much. asked it just for a bug description and that one i then gave to `amp`, whose solution I then again reviewed with pi+gpt5.5, to me this change now looks reasonable. Also told amp to "use tmux, start it with the -w flag so that i can test it using codiff itself. That was fun, but it only created 1 card whose description wasn't too helpfull. I'll try again soon with something else.

If you end up just prompting yourself and not merging this, no worries, no hard feelings :) 

amp's PR description below:

## Summary
- resolve the Codex CLI with `/bin/zsh -lc 'command -v codex'` before falling back to hardcoded paths
- keep `CODIFF_CODEX_PATH` as the explicit override
- cache the async shell lookup so the Electron main process does not block or repeat shell startup for every Codex launch

## Testing
- `pnpm exec vp check --fix`
- `pnpm exec vp build`
- `pnpm codiff -w`

### DIFF
--- a/electron/codex.cjs
+++ b/electron/codex.cjs
@@ -1,9 +1,10 @@
 // @ts-check
 
-const { spawn } = require('node:child_process');
+const { execFile, spawn } = require('node:child_process');
 const { existsSync, promises: fs } = require('node:fs');
 const { tmpdir } = require('node:os');
 const { join } = require('node:path');
+const { promisify } = require('node:util');
 
 const CODEX_TIMEOUT_MS = 45_000;
 const DEFAULT_OPENAI_MODEL = 'gpt-5.3-codex-spark';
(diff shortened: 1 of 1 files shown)


cpojer:
Thank you so much for the investigation, this is really helpful. In one way, it is user error, but in another, it can happen if you have various versions of node, package managers, and node version managers. I'm not merging your PR (calling the login shell is sketchy), but I'm shipping an improvement to provide more clarity and an actionable error message to tell users how to get out of such a situation.

It's on main now, will make a release tomorrow.

cpojer:
Commit: https://github.com/nkzw-tech/codiff/commit/af75f927b4564eb275b23f9a63e63ce22530722b

Again, really appreciate your investigation.

PR #20, opened 2026-05-21
fix: restore last repository on app launch

lyzno1:
## Summary
- remember the last successfully opened repository in preferences
- restore it on plain app launches without overriding explicit CLI path/source launches
- add tests for restore and explicit launch boundaries

### DIFF
--- a/electron/__tests__/command-line.test.ts
+++ b/electron/__tests__/command-line.test.ts
@@ -1,18 +1,37 @@
+import { mkdtemp, rm } from 'node:fs/promises';
 import { createRequire } from 'node:module';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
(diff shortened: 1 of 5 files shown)


cpojer:
Why do you need this? Do you start Codiff from the dock/spotlight?

lyzno1:
> Why do you need this? Do you start Codiff from the dock/spotlight?

Yes, sometimes I open the app from Spotlight. In that flow Codiff always starts with an empty/no-repository window. The more annoying part is that if I then use File → Open Folder, it opens the repository in a new window, leaving the original empty window behind.

So this PR is mainly trying to make the existing direct app launch flow less awkward. The CLI flow should not be affected because CLI launches still pass explicit repository context.

cpojer (approved):
Awesome, makes sense, thank you!

PR #21, opened 2026-05-21
fix: resolve relative repository paths

lyzno1:
## Summary

- Resolve relative repository path arguments in the packaged terminal helper before launching Electron.
- Cover dot, nested relative paths, and walkthrough launches with helper-level tests.


cpojer (approved):
Awesome, thanks!

PR #23, opened 2026-05-22
Make walkthrough groups collapsible

kyleparrott:
## Summary

Adds collapsible walkthrough groups in the sidebar so reviewers can focus on one review strategy at a time.

I've been enjoying codiff so far, especially the walkthroughs.  This is a small quality-of-life tweak that made walkthroughs on larger PRs a bit easier to work with.

image

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -750,16 +750,58 @@
 }
 
 .walkthrough-group-header {
+  -webkit-app-region: no-drag;
+  background: transparent;
+  border: 0;
+  border-radius: 8px;
   color: var(--text);
+  cursor: pointer;
   display: flex;
   flex-direction: column;
   font: 700 11px/1.25 var(--font-sans);
   gap: 2px;
   padding: 5px 6px 3px;
+  text-align: left;
+  width: 100%;
 }
 
-.walkthrough-group-header span {
+.walkthrough-group-header:hover {
+  background: rgb(127 127 127 / 0.08);
+}
+
+.walkthrough-group-title-row {
+  align-items: center;
+  display: grid;
+  gap: 6px;
+  grid-template-columns: 14px minmax(0, 1fr) max-content;
+  min-width: 0;
+}
+
+.walkthrough-group-chevron {
+  color: var(--muted);
+  transform: rotate(0deg);
+  transition: transform 160ms ease;
+}
+
+.walkthrough-group.collapsed .walkthrough-group-chevron {
+  transform: rotate(-90deg);
+}
+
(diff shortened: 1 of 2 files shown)


cpojer:
Awesome! Thank you.

Sorry you missed the train for 0.6.0 just by a bit, and I wasn't going to make another release today. You ok using your local checkout for now? :D

kyleparrott:
Sure, that's no problem! This can definitely wait for the next release.

cpojer:
Just pushed another change to improve the styling of the walkthrough section a bit.

It's better, but I'm still not happy with it. I think there must be a 10x improvement possible to make the walkthrough section much more impactful and less sloppy.

kyleparrott:
@cpojer I think you’re right. My read is that the current walkthrough UI is doing a few things at once - showing a full list of all the files changed, a suggested order for the review, and some conceptual explanation.

One direction I’ve been thinking about is whether it could make sense to structure the walkthrough experience and UI around “concerns” or functional slices that changed in the PR/branch. For example, a large PR may include the implementation for api endpoints, UI changes, and backend persistence for a hypothetical feature. A reviewer may want to see each of these slices and its related diff in isolation during a walkthrough, which could allow for a more focused UI.

This would be a pretty ambitious and sizable change - it could make sense as a separate mode for walkthroughs. It'd likely need some prototyping to see how well it works in practice and how useful it is across the range of PR sizes. My guess is this would be most useful for larger PRs that the reviewer didn’t author themselves, which is just one of the many use cases for Codiff.

I'm curious what you think.

cpojer:
Personally, I would like to keep Codiff simplistic. The goal of the app is to make it easier to review local code changes. Having two modes for walkthroughs or adding layers of complexity that are hard to use does not vibe with the goals of the project. With that in mind, I think you are exploring into the right direction. In the end, I *do* think sorting and explaining the changes well is the best area for improvement, but I'm definitely open to prototypes – with the intent of replacing the current walkthrough mode with something that is better, rather than adding a new mode.

cpojer:
I'm currently experimenting with a `$codiff` skill that uses the current session to produce walkthroughs, and quality appears higher compared to before.

kyleparrott:
Sure, that framing makes sense. I’ll see what I can find when I have some time. I’m looking forward to trying out that skill too - that’s a great idea.

cpojer:
I will ship this today: https://github.com/nkzw-tech/codiff/commit/631719cc0306b24100833924b9f9f04b3ba1223d

PR #25, opened 2026-05-22
Show proper history when diffing a pull-request

bryceosterhaus:
`codiff https://github.com/nkzw-tech/codiff/pull/14`

I'm not totally sold on the headers like "In this PR", but I do kind of like it. I also like that I can select the top of the history and diff all the files in the PR

**Before:**
Screenshot 2026-05-22 at 11 23 52 AM

**After:**
Screenshot 2026-05-22 at 11 31 48 AM

Note: used gpt 5.5 for this

### DIFF
--- a/electron/git-state.cjs
+++ b/electron/git-state.cjs
@@ -7,6 +7,8 @@ const {
   readCommitState,
 } = require('./git-state/commit.cjs');
 const {
+  listPullRequestHistory,
+  normalizeGitHubPullRequestCommit,
   normalizeGitHubReviewComment,
   normalizePullRequestComment,
   parseGitHubPullRequestUrl,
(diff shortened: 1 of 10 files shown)


cpojer:
Is the idea that instead of looking at the whole PR, you can step through individual commits?

I'm not opposed since it is easy to add and doesn't hurt, but I'm trying to figure out the use case. In 18 years (minus 1 month) on GitHub, I can count the times I looked at individual commits by one hand.

bryceosterhaus:
Yeah the purpose would be align with reviewing a PR commit by commit. I actually do this quite a bit. Definitely depends per person though and the teams you might work on though. But generally I like to review per-commit, if they are cohesive units of work, to see the mental flow of how something was composed together. 

Regardless though, I think the PR view should show the proper history instead of the local repo history.

cpojer:
Ok, can you move the titles to the left and have the llm apply the same headline styling as in walkthroughs?

bryceosterhaus:
Screenshot 2026-05-22 at 12 20 56 PM

Updated titles in most recent commit

cpojer:
Awesome, much better, thank you!

PR #27, opened 2026-05-22
Add --help and --version flags to the CLI

bkrem:
## Context

Thank you for building codiff, it's already become part of my standard workflow ❤️ 

A current struggle for me as a user is that there isn't a `--help` command or similar on the CLI, making discoverability difficult beyond going to the repo's README and checking. Goal was to implement this in a straightforward and maintainable way.

E.g. I wasn't even aware you could do `codiff pr <pr-number>` until I saw this tweet: https://x.com/cnakazawa/status/2057208130229223886

## Summary

- Adds `--help` / `-h` and `--version` / `-v` to both CLI entry points
  (`bin/codiff.js` and `bin/codiff-app`)
- Flag definitions live in a single structured array in `arguments.js` so
  adding future flags only requires one addition
- Prints usage info and exits before launching Electron
- Includes unit and integration tests for the new flags

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,12 @@ codiff -w
 codiff -w a1b2c3d
 ```
 
+Show all available options:
+
+```bash
+codiff --help
+```
+
 Launching Codiff in multiple repositories opens a separate native window for each repository.
 
 ## Command Bar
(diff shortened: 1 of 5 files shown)


cpojer:
Hey, thanks for the kind words, and for sending a PR. Totally makes sense, let's ship this :)

> I wasn't even aware you could do codiff pr <pr-number> until I saw this tweet

You figured out my twitter influencer strategy!

PR #31, opened 2026-05-24
fix: persist window dimensions across sessions

bkrem:
## Context
Offering another small QoL improvement here: currently window dimensions are not persisted in the app, which means the user needs to resize windows manually (e.g. to full screen, half width, etc) from the default dimensions everytime `codiff` is called fresh. This breaks up flow state when wanting to jump straight into the code diff.

## Summary
- Save window size, position, maximized, and fullscreen state to `~/.codiff/window-state.json` on close
- Restore saved state on next launch, with off-screen validation (100px overlap threshold against all displays)
- Fall back to centered 86%-of-screen defaults when no state exists, file is corrupt, or saved position is off-screen
- Track pre-maximized/pre-fullscreen bounds so closing while maximized preserves the user's preferred windowed dimensions
- 12 new unit tests in `window-state.test.ts` covering read/write/validation logic
- Tested manually on macOS across resize, maximize, fullscreen, and corrupt/missing state file scenarios

## Demo
https://github.com/user-attachments/assets/599f49a4-a1fa-4c4b-a295-8c594ae26699

### DIFF
--- a/electron/__tests__/window-state.test.ts
+++ b/electron/__tests__/window-state.test.ts
@@ -1,0 +1,169 @@
+import { mkdtemp, rm, writeFile } from 'node:fs/promises';
+import { createRequire } from 'node:module';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { expect, test } from 'vite-plus/test';
+
+const require = createRequire(import.meta.url);
+const { readWindowState, validateWindowStateOnScreen, writeWindowState } =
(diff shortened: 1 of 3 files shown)


cpojer:
Awesome, thank you!

cpojer:
Published in 0.8.0. Thank you!

bkrem:
> Published in 0.8.0. Thank you!

Likewise! 🙏 🚀

PR #32, opened 2026-05-25
feat: hide resolved and outdated review comments

etrepum:
On an active PR it's common for review threads to be resolved or anchored to lines that have since changed. GitHub's review UI tucks both away, and this implements the same filtering for codiff.

Resolved threads are filtered out entirely. Outdated comments are hidden by default behind a persisted "show outdated" setting, toggled from the View menu and the "Toggle Outdated Comments" command.

Implements #30


cpojer (approved):
Thank you!

cpojer:
Shipped in 0.8.0.

PR #33, opened 2026-05-26
Add unified diff layout setting

gumadeiras:
## Summary

Adds a diff layout setting so users can switch between split and unified diffs.

Users can change the layout from:
- View menu
- command bar: Toggle Diff Layout
- config: `settings.diffStyle`

The command bar action shows the target layout, for example `Switch to Unified`.

## Verification

- `vp check --fix`
- `vp build`
- `vp run --no-cache test:all`


cpojer (approved):

cpojer:
Awesome! Thank you so much. This missed today's release train, but will be in the next release :)

PR #34, opened 2026-05-26
Fallback when avatar images fail to load

gumadeiras:
## Summary

Adds a fallback for broken avatar images.

If a Gravatar or GitHub avatar URL fails to load, Codiff now shows the existing initial fallback instead of a broken image icon.

## Verification

- `vp check --fix`
- `vp build`
- `vp run --no-cache test:all`


cpojer (approved):
Nice, thanks!

PR #38, opened 2026-06-01
Add word wrap setting for diffs

gumadeiras:
Adds a word wrap setting for the diff view.

This makes docs and other long lines easier to read without horizontal scrolling.

Users can toggle it from `View > Word Wrap`, use `Toggle Word Wrap` in the command bar, or set `settings.wordWrap` in `~/.codiff/codiff.jsonc`.

The setting is saved through the existing config flow and passed to the diff renderer as `overflow: 'wrap'` when enabled.

The renderer now derives preferences from the settings object directly instead of copying each setting field by hand.

Validation:
- `vp check --fix`
- `vp build`

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,7 @@ is running so changes apply to open windows.
     "openAIModel": "gpt-5.3-codex-spark",
     "showWhitespace": false,
     "theme": "system",
+    "wordWrap": false,
   },
   "keymap": {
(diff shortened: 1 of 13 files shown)


gumadeiras:
on the copilot comment: i believe diffs pkg takes care of that automatically and it should not be a problem (it force updates the layout when overflow changes)

cpojer:
Thank you!

PR #39, opened 2026-06-02
Expand unmodified context in pull request diffs

etrepum:
Pull request diffs only rendered the GitHub-provided patch -- a partial diff with no way to reveal the surrounding unmodified lines. This brings the same GitHub-style "expand unmodified lines" context expansion that the working tree and commits already have to pull requests.

How it works:

- Each pull request file's base and head contents are read from the local refs (the pull request is already fetched), diffing the head against the merge base of the base branch and head -- the same comparison GitHub uses -- so line numbers and existing review comments stay aligned. With both sides present the diff is recomputed locally and the expand controls light up.

- Contents are loaded up front, batched, when the pull request state is read, so every diff renders in its final collapsed layout immediately and never shifts as you scroll or as a file's contents settle. Files past the eager size limit stay patch-only and load on demand.

- Loaded diffs are parsed from the existing patch with the full contents attached rather than recomputed from scratch, so the collapsed layout matches the patch-only view, and the collapse threshold is aligned with the context git/GitHub patches emit (-U3). As a result the collapsed view shows a small, consistent amount of context (expand for more) and only expanding a section yourself changes its height.

On-demand section loading for the working tree and commits now runs from a visibility-based loader that loads whatever is on screen once scrolling settles.

Includes tests for pull request section building (full-content vs patch fallback) and for resolving the diff against the merge base.

### DIFF
--- a/electron/git-state.cjs
+++ b/electron/git-state.cjs
@@ -11,6 +11,7 @@ const {
 const {
   collectResolvedReviewCommentIds,
   createPullRequestHistoryFetchRefspecs,
+  createPullRequestSection,
   getPullRequestHeadImageSource,
   listPullRequestHistory,
   normalizeGitHubPullRequestCommit,
@@ -19,6 +20,7 @@ const {
   parseGitHubPullRequestUrl,
   readPullRequestImageContent,
   readPullRequestState,
+  resolvePullRequestContentRefs,
   selectUnresolvedReviewComments,
   submitPullRequestComment,
   submitPullRequestReview,
@@ -79,6 +81,7 @@ const readDiffImageContent = (launchPath, request) =>
 module.exports = {
   collectResolvedReviewCommentIds,
   createPullRequestHistoryFetchRefspecs,
+  createPullRequestSection,
   getPullRequestHeadImageSource,
   listRepositoryHistory: readRepositoryHistory,
   normalizeGitHubPullRequestCommit,
@@ -96,6 +99,7 @@ module.exports = {
   readPullRequestState,
   readRepositoryState,
   readWorkingTreeState,
+  resolvePullRequestContentRefs,
   submitPullRequestComment,
   submitPullRequestReview,
   validateRepositoryPath,
--- a/electron/git-state/pull-request.cjs
+++ b/electron/git-state/pull-request.cjs
@@ -10,6 +10,8 @@ const {
   getImageMimeType,
   git,
   gitOrEmpty,
+  readGitFile,
+  summarizeContent,
   validateRepositoryPath,
 } = require('./common.cjs');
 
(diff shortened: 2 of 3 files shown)


cpojer:
Clanker says:

Findings

  - Medium: electron/git-state/pull-request.cjs:633 only refreshes cached PR refs when the base
    ref is missing or the head SHA changed. It does not verify metadata.base.sha against the
    local refs/codiff/pull-requests/<n>/base. If a PR is retargeted, the base branch is force-
    pushed, or the base ref otherwise changes to a different merge base, Codiff can render
    expandable PR diffs against stale base contents. I would refetch when localBase !==
    metadata.base.sha when available, or always refresh the PR base/head refs before resolving
    content refs.

  - Low/Medium: src/lib/diff.ts:279 now prefers processFile(section.patch, { oldFile,
    newFile }) for loaded full-content diffs. That path bypasses ignoreWhitespace: !
    showWhitespace, so loaded PR diffs no longer honor the default “hide whitespace-only
    changes” behavior that commit/working-tree full-content diffs use. The fallback
    parseDiffFromFile(... ignoreWhitespace ...) only runs when there is no patch. This can make
    whitespace-only PR changes remain visible and affect filtering/search paths that use
    parseSectionDiffWithOptions.

  Merge Recommendation
  I would not merge as-is. The stale base-ref issue is enough to request a fix before merging;
  the whitespace behavior should also be fixed or explicitly accepted as a product change.

cpojer:
Clanker say: I would not merge PR #39 as-is.

  Findings

  - High: electron/git-state/pull-request.cjs:671 detects binary files by running /Binary
    files .* differ/ across the entire patch text. This PR adds that literal text in source/
    tests, so normal text files are misclassified as binary. I confirmed with
    readRepositoryState on PR #39: electron/git-state/pull-request.cjs and src/__tests__/git-
    state.test.ts came back as loadState: 'binary' with no oldFile/newFile, so Codiff would
    hide their text diffs. The check should only match actual binary diff metadata lines, e.g.
    an anchored diff line, not arbitrary patch content.

  - Medium: patch-only PR sections are treated as loadable by shouldLoadDiffSectionContents,
    but src/App.tsx:184 blocks loading for pull-request sources, and electron/git-state.cjs:66
    has no PR-specific section-content loader anyway. If PR content refs cannot be resolved, or
    a file falls back to patch-only, there is no working path to load expandable context later.
    
 tbh I would prefer if you used Codex 5.5 🥲

etrepum:
Alright, I'll apply for Codex for open source, but in the meantime what I have is Claude. Here's what they had to say about that (re: binary matching) 😆 

> One thing to flag for transparency: the same unanchored regex pattern is pre-existing in common.cjs, working-tree.cjs, and commit.cjs. It's not in this PR's diff so the reviewer didn't surface it, but it's the same latent bug for working-tree/commit diffs. I kept this PR scoped to the flagged file — happy to anchor those too (here or as a separate change) if you'd like the codebase consistent.
>
> And — fair enough on the tooling preference 🙂. I'll keep tightening the loop and being precise about exactly these kinds of edge cases.

cpojer:
> And — fair enough on the tooling preference 🙂. I'll keep tightening the loop and being precise about exactly these kinds of edge cases.

Love how it says that and the instantly forgets about it, obviously.

cpojer:
Thank you!

PR #40, opened 2026-06-02
Preserve file selection across reloads

gumadeiras:
## Summary

This keeps the selected source and file stable when Codiff reloads. Maintains the currently selected file in the sidebar and in the diffs view. It also highlights only the git status glyph for files that changed since the last refresh, so we can quickly see which files changed since the last reload without adding extra UI (see orange M in the example).

image

## Why

This makes reloads feel like a refresh, not like relaunching codiff. The implementation stores a small one-shot reload snapshot, restores the selected file only when it still matches the same source/root, and compares file fingerprints to mark changed files after reload.

## Testing

- `pnpm exec vp check --fix`
- `env GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=commit.gpgsign GIT_CONFIG_VALUE_0=false pnpm exec vp test`
- `pnpm exec vp build`


cpojer:
Why is it a draft?

PR #42, opened 2026-06-02
Add a commit details panel

gumadeiras:
## Summary

Adds a commit details panel for commit reviews from the History tab.

The panel only appears when viewing a committed change. It does not show for local uncommitted changes.

The panel shows the commit message, author, committer, refs, parents, signature, trailers, file stats, and the files changed in the commit.

Changed files in the panel are clickable, so you can jump straight to that file in the diff. Reloading keeps the restored scroll position instant, while clicking a file scrolls smoothly to it (can change it to instant if you prefer).

image

I added the commit details panel as a fake item in the diff list. This keeps the diff view simple; the panel scrolls with the files, reloads restore the right position, and clicking a file can use the same navigation path as the rest of the diff.

## Validation

- `vp check --fix`
- `vp build`
- `vp test src/__tests__/App-render.test.tsx`

### DIFF
--- a/electron/git-state/commit-metadata.cjs
+++ b/electron/git-state/commit-metadata.cjs
@@ -1,0 +1,252 @@
+// @ts-check
+
+const { fileSort, getGravatarHash, git, gitBufferWithInput, gitOrEmpty } = require('./common.cjs');
+
+/**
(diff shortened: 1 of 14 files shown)


cpojer:
Made a bunch of styling improvements and going to ship this as 1.0.0 now. Thank you for all your contributions!

PR #45, opened 2026-06-03
Fix app icon sizing on macOS and Windows

gumadeiras:
This updates the generated app icon assets so they fit each platform better.

- macOS icon now uses an optical inset, so it no longer looks oversized in the Dock.
- Windows icon now includes multiple standard sizes: 16, 24, 32, 48, and 256.
- Linux PNG remains full-size, so the macOS sizing change does not shrink Linux icons.

Before:
image

After:
image

Validation:
- Ran `vp check --fix`
- Ran `vp build`

`vp build` still reports the existing Vite chunk-size warning.


cpojer:
You are the man!

cpojer (approved):

cpojer:
I'm so confused, I thought this worked (and you have screenshots to prove it), but I just installed Codiff on another computer and the icon is still the old small one on Tahoe… Do you mind taking another look?

gumadeiras:
> I'm so confused, I thought this worked (and you have screenshots to prove it), but I just installed Codiff on another computer and the icon is still the old small one on Tahoe… Do you mind taking another look?

small as it's in [squircle jail](https://www.heise.de/en/news/Icons-in-macOS-26-Fighting-the-Squircle-Prison-11075561.html)? I suspect it's because of how Tahoe deals with icons; I'll check on this and put up another PR, will probably need to use compiled icons for macOS instead of a generic .icon
I'm on Sequoia still but will set up a Tahoe VM to verify

PR #48, opened 2026-06-04
Add Claude Code as a selectable agent backend

orta:
Adds support for Claude Code as an agent 

image

Basically abstracts the agent into a interface and both cc and codex operate with it, robot opinion below:

----

Codiff previously hardcoded the OpenAI Codex CLI for walkthroughs and inline review assistance. This introduces a small agent-backend abstraction and a second backend (Claude Code), selectable per-config, via the app menu, or with a `--agent` flag.

What changed:
- electron/agent.cjs resolves a backend ('codex' | 'claude') into a uniform interface: run(), models, normalizeModel, notFoundCode/isNotFoundError, readSessionContext, and a skill descriptor. walkthrough.cjs/review-assist.cjs now take an `agent` instead of calling runCodex directly.
- electron/claude.cjs runs `claude -p --output-format json --json-schema … --tools "" --add-dir <repo> --no-session-persistence`, reads the validated `structured_output` field, and falls back across models / detects login.
- electron/claude-session-context.cjs parses ~/.claude/projects/**/<id>.jsonl.
- New config: settings.agentBackend (default "codex") + settings.claudeModel.
- A `/codiff` Claude Code skill (claude/skills/codiff) mirrors the Codex skill.
- Renderer labels (Ask <agent>, reply header, unavailable panel, skill button) follow the active backend.

Trade-offs and decisions:
- Default stays Codex. This is additive, not a migration: existing configs and the Codex path are untouched, so the change is opt-in.
- Per-backend model settings (openAIModel + claudeModel) instead of one shared field. Keeps existing Codex configs working verbatim at the cost of a second setting and a small amount of menu/config branching.
- Structured output relies on Claude Code's `--json-schema` flag returning a validated `structured_output` envelope (verified against claude v2.1.x). This is newer/less battle-tested than Codex's --output-schema; if a model ignores the schema we fall back to parsing the `result` text.
- Claude rides the user's existing OAuth/subscription login (no --bare, which would force ANTHROPIC_API_KEY). Simpler setup, but it means a not-logged-in CLI surfaces as "unavailable" rather than auto-authenticating.
- Default Claude model is Sonnet 4.6 (not Opus): walkthroughs run under a ~90s budget, so latency/cost matter more than peak quality for review ordering.
- `--tools ""` disables tool use so Claude answers only from the diff digest, matching Codex's read-only sandbox. Claude may still load CLAUDE.md/git status as ambient context (slightly more input tokens); the prompt tells it to treat the digest as the source of truth.
- Internal renderer identifiers (codexReply, askCodex) were left as-is to keep the diff reviewable; only user-facing strings became agent-aware. A full rename is deferred to avoid churn.
- Kept the flatter electron/ layout (codex.cjs in place, claude.cjs alongside) rather than relocating into electron/agents/, so existing test imports and the codex-skill.cjs wrapper stay stable.

Tests: added claude.cjs, claude-session-context.cjs, agent registry, and CLI / skill-launcher coverage; full suite green (199 tests).


cpojer:
Awesome.

Can you please use the actual Claude logo and make it look pretty similar to how the Codex logo is shown like a regular profile pic? I'm afraid shipping it as is will hurt your high-quality UI brand.

orta:
that is very reasonable

orta:
image

That is in!

I'll try to give this a whirl on a few puzzmo PRs before I look again, but from a quick play around - I would be interested in adding `shift + ?` support to show the key bindings from your settings inline, and then adding commands for jumping between hunks in the diff (probably ctrl + up / down)

cpojer:
The codex logo looks like an avatar with a background color and squircles applied, can we not use the avatar component that's already in the codebase?

Happy to add full support for keyboard shortcuts for any action. Just put it in this PR if you like, I don't care.

orta:
The claude icon is just the squiggle, it is in that same squircles avatar view it just has transparency 

image

I tried adding a bit of the orange to the bg, but TBH, it looks meh

image

I've added a keybindings panel, and a shortcut to change the word wrap

orta:
Maybe I can find an example of the icon in orange with a white foreground

PR #52, opened 2026-06-05
Walkthrough 2

orta:
Still probably some testing required but, this fixes #49 

My theory based on the work we have done at Artsy is that a guided PR can be built where there is a solid upfront effort by somebody to create a narrative around the diff.  e.g. https://github.com/artsy/eigen/pull/4055

So, we have the agent try to make a 'narrative arc' for the change, using its existing conversational context and the actual diff itself. 

This means having each hunk / file change be marked in some way in a linear format, but also that the entire diff does not necessarily have to be the narrative arc. So I think of it as their is:

- Files in the core arc for the diff (e.g. in this case adding the json handler, updating the view layer)
- Files which changed due to that arc (e.g. tests, snapshots, lockfiles)
- Files with weren't a part of it, but still exist (css changes, skill changes, icons etc)

From those we can do something cool at the view layer to make a guided walkthrough of the diff!

---

Roughly speaking, this adds a new mode where a skill can generate an intermediary JSON file which codiff can be opened with. Here is the one (with some cuts for readability) for this PR: 

````jsonc
{
  "$schema": "https://raw.githubusercontent.com/nkzw-tech/codiff/main/src/walkthrough/narrative-walkthrough.schema.json",
  "version": 2,
  "kind": "narrative",
  "agent": "claude",
  "generatedAt": "2026-06-05T13:50:00.000Z",
  "title": "Narrative walkthrough — data model + the C/D interface",
  "focus": "A new story-shaped walkthrough for Codiff: order-independent segments anchored into the live diff, multiple reading orders over them, a validator that repairs the document, a /walkthrough skill that authors it, and the hybrid arc + results-first UI that renders it — using the real diff engine for each stop.",
  "meta": "19 stops · 7 chapters · the feature, reviewing itself",
  "repo": { "root": "/home/orta/dev/oss/codiff", "branch": "main" },
  "source": { "type": "working-tree" },
  "defaultOrder": "keys",
  "segments": [
    {
      "id": "types",
      "path": "src/types.ts",
      "status": "modified",
      "granularity": "hunk",
      "added": 131,
      "deleted": 0,
      "title": "The NarrativeWalkthrough contract",
      "anchor": {
        "display": "src/types.ts:252",
        "sectionKind": "unstaged",
        "side": "additions",
        "startLine": 252,
        "endLine": 360
      }
    },
    {
      "id": "validator",
      "path": "electron/narrative-walkthrough.cjs",
      "status": "untracked",
      "granularity": "file",
      "added": 531,
      "deleted": 0,
      "title": "The trust boundary",
      "anchor": { "display": "electron/narrative-walkthrough.cjs (new)", "side": "additions" }
    },
   ],
  // ...
  
   "orders": [
    {
      "id": "keys",
      "label": "Key changes first",
      "tagline": "Build order: contract → renderer.",
      "restLabel": "Not in the arc",
      "restBlurb": "Glue lines, the Codex mirrors, and small fixtures — skim only.",
      "phases": [
        {
          "id": "contract",
          "n": 1,
          "title": "The contract",
          "icon": "path",
          "blurb": "The shape everything agrees on."
        },
        {
          "id": "boundary",
          "n": 2,
          "title": "The trust boundary",
          "icon": "wrench",
          "blurb": "Repairing the document against the diff."
        },
       // ...
    ] ,

  "context": {
    "version": 1,
    "objective": "Build a narrative walkthrough for Codiff: a flexible JSON data model plus the hybrid (C) arc + results-first (D) interface that renders it, generated by a /walkthrough skill and carrying the prior conversation for in-app Q&A.",
    "decisions": [
      "Stops reference into the live diff (file + section + line range) rather than embedding diff rows.",
      "The /walkthrough skill writes the JSON itself, since the agent already holds the conversation.",
      "Emit all reading orders in one document with an agent-chosen defaultOrder; the UI toggles between them.",
      "Separate order-independent segments from orders so results-first and key-changes-first are two views of one dataset.",
      "Render each stop with the real @pierre/diffs ReviewCodeView, scoped to one file, so inline comments reuse the existing flow."
    ],
    "constraints": [
      "Keep the v1 file-ordering Walkthrough untouched and additive.",
      "Defer the All-files flat switch-out and hunk-only slicing to a follow-up."
    ],
    "validation": [
      "vp check passes (types + lint).",
      "vp test passes: 217 tests including the normalizer and view-model suites.",
      "vp build bundles the renderer."
    ],
    "implementationSummary": "Phase 1 added the NarrativeWalkthrough v2 types, the normalizer/validator, a committed schema, the /walkthrough skill (claude + codex), and --walkthrough-file → getNarrativeWalkthrough wiring. Phase 2 added the pure view-model, the shared navigation hook, the hybrid arc + focus view, the narrative sidebar ToC, ported CSS, and the App/Sidebar swaps — each stop rendered by the real diff engine.",
    "source": { "generatedAt": "2026-06-05T13:50:00.000Z", "type": "claude-session" }
  }
}

````

this file is then passed in to codiff as the metadata surrounding the diff it should be looking at.

My goal with the intermediary file is sorta twofold: the context for generating it is already whereever you are chatting with the agent and Claude Code's `-p` is not a part of the subscription and that  makes using it inside codiff have a real $ cost each time. I would avoid using it except when necessary. 

---

There are two modes for the walkthrough:

The core first, and then the rest

Screenshot_2026-06-05_13-55-42

Tests / artefacts, the core, the rest

Screenshot_2026-06-05_14-18-08

The rest is treated as a final section of files which are part of the diff but not the core parts
 
Screenshot_2026-06-05_14-22-35


orta:
Future ideas:

- Add a watcher to the JSON file and allow the agent to send updates

- ~~The skill should probably get most of its info from the installed version of codiff (e.g. the skill you install tells your agent how to get info (like the shape of the JSON, recommendations for how to structure the files) from running `codiff narrative --skill` or something. That keeps it in sync and doesn't require skill changes in user-land~~

- My personal flow would be to add this into my commiting pipeline, where you read all the changes and then at the end are presented with a 'now what do you want to commit' type of page with an example base commit message and you write the human title for it

cpojer (approved):

cpojer:
Amazing @orta, thank you.

Merge first, ask questions later.

PR #59, opened 2026-06-08
Add local code font selection for diff rendering

vinitkumar:
## Summary

   Adds a local font picker for code diffs.

   Users can choose an installed font family, reset back to Fira Code, and see the
   current selection from the action bar, command bar, or View menu.

   ## Implementation

   - added persisted `settings.codeFontFamily`
   - used `window.queryLocalFonts()` to list installed local fonts
   - applied the selected font only to diff/code rendering via `--font-diff-mono`
   - added picker entry points in:
     - `View > Diff > Choose Code Font…`
     - command bar
     - bottom-right action bar pill
   - updated docs and test mocks

   ## Testing

   - `pnpm exec vp check --fix`
   - `pnpm exec vp test src/__tests__/config-defaults.test.ts src/__tests__/App-render.test.tsx`
   - `pnpm exec vp build`

## Screenshots

Screenshot 2026-06-08 at 10 41 44
Screenshot 2026-06-08 at 10 42 04


cpojer:
Looks absolutely horrible tbh. I shipped a new version of this that adds font size controls to the menu (including keyboard shortcuts like in other apps), adds it to the command palette, and you can set the font size through the config option – we really don't need to add a persistent button and overlay UI for this.

Will ship in 1.3.0 later this week.

PR #61, opened 2026-06-08
Recommend adding code blocks into the sections for describing upcoming parts of the walkthrough

orta:
Passages like this are just begging for code references to be a different font

image


cpojer:
Wait, we already have a markdown renderer. Can we tell it to hand us markdown that we render? Then code blocks will look nice, too.

orta:
Yeah, I double checked, those are already markdown renderers

PR #67, opened 2026-06-08
Use hunk-based walkthroughs for focused review + single renderer

gumadeiras:
## Summary

This moves narrative walkthroughs from agent-authored file anchors to deterministic diff hunk IDs.
My experience has been getting mostly full file diffs or near full file diffs with no explicit order for information flow in the code; just order for reviewing files. When information flow within/across files matters, this makes reviewing slow.  Switching to a hunk-based approach gives much better granularity and review order in my tests. When full file diffs are needed, they happen naturally.

The agent now provides review judgment: which hunks matter, how they should be grouped, what order to read them in, and the prose for each step.

Another problem I noticed is that information the agent provides can mismatch what is displayed, especially lines added/removed and other computed metadata. This changes so anything that can be computed deterministically is not asked of the agent. Codiff computes the diff facts from the live diff: paths, statuses, anchors, line counts, labels, focused patches, and uncovered Support items (rest -> Support; any hunks that were not assigned to be explicitly reviewed are shown there).

Another change is making walkthrough renders through a single review surface, avoiding many separate diff views. For large diffs I was getting a lot of perf problems and slow scrolling/typing because each diff item was an independently rendered unit.

I'm still testing and will do some cleaning up, but the core of the change is there and seems to be working well. I have a few more things to clean up and test, but wanted to get this out for early feedback. Also happy to just close this or separate out just the single renderer change since that's still useful perf improvement.

image

image

image

## What changed

- Added v4 walkthrough JSON based on `hunkIds`
- Render walkthrough stops as focused hunk blocks in one review surface
- Preserve authored review order while grouping adjacent hunks from the same file
- Add Support mode for generated, mechanical, omitted, and uncovered hunks
- Remove agent-authored inline comments from the walkthrough schema
- Key viewed/collapsed state by review identity so repeated file views work correctly

## Refactored

- Removed walkthrough order preference, menu, IPC, and config schema; walkthroughs now use one authored order
- Extracted shared hunk diff utilities into `shared/narrative-walkthrough-diff.cjs`
- Extracted Electron walkthrough schema and constants into `electron/narrative-walkthrough-schema.cjs`
- Extracted review identity and review command target helpers
- Extracted test helpers for fixtures, React rendering, CLI fake commands, and the CodeView mock
- Removed unused walkthrough totals and block note override plumbing

## Validation

- `vp check --fix`
- `vp test`
- `vp build`

### DIFF
--- a/bin/walkthrough-guide.md
+++ b/bin/walkthrough-guide.md
@@ -20,11 +20,13 @@ choose.
   `"Tests"`, `"Docs"`, `"Runtime"`, `"Cleanup"`.
 - **`chapters[].stops[]`** — the main review path. Use 3-6 stops for small changes, 5-9 for
   medium changes, and 7-12 for large changes. Never exceed 14. A stop should represent one
-  review idea and can include up to 3 ordered `hunkIds`.
+  review idea and can include up to 14 ordered `hunkIds`.
 - **`hunkIds[]`** — deterministic hunk ids copied from the repository digest, in the exact order
   Codiff should render them. Default to one hunk id per stop or support item. Use multiple ids
   only when those hunks must be read together; cross-file and out-of-line order is allowed when it
-  improves the review path.
+  improves the review path. Some ids are synthetic hunks for binary, deferred, metadata-only, or
+  otherwise non-textual changes; treat them like normal hunk ids and explain why that review unit
+  matters.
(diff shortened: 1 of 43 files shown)


cpojer:
Can you rebase this one?

gumadeiras:
> Can you rebase this one?

done!

cpojer:
Awesome, let's merge and improve as we go forward.

orta:
nice work

PR #68, opened 2026-06-08
Mute viewed files in the tree sidebar

gumadeiras:
This updates the Tree tab so files marked as `Viewed` are easier to distinguish. Viewed files now mute their row content, including the file icon, path, line count, and git status.

If the fingerprint for a file changes, Viewed gets reset. Also made the viewed toggle logic shared between the keyboard command and Viewed button so they don't drift.

Tests cover current viewed files, stale viewed entries, and the muted tree row selectors.

image

image

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -388,6 +388,42 @@ export default function App() {
     }, 1200);
   }, []);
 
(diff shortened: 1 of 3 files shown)


cpojer:
Niiice.

PR #69, opened 2026-06-08
Keep large walkthroughs responsive

ericchernuka:
Large walkthroughs could get painfully slow when the diff had a lot of files. The bad path was mostly renderer work: too many live diff views were mounted in walkthrough mode, and typing a comment pushed every keystroke back through App state.

This keeps walkthrough mode to one active step at a time. The arc arrows, j/k, and the bottom Next button swap the current step. The active step uses one bounded `ReviewCodeView` with the files for that step, so it stays closer to the main review surface instead of keeping a stack of diff views alive.

Comment typing now stays local until blur or action. Blur still persists the draft. Ask and Comment also pass the visible draft body into App, which is intentional: while the textarea owns the draft, `comment.body` can lag behind what the reviewer is looking at.

Benchmarks from the large branch walkthrough used for profiling:

| Scenario | Before | After | Improvement |
| --- | ---: | ---: | ---: |
| Walkthrough Next navigation | 1.75s-4.13s | 56.8ms-119ms | ~93-97% faster |
| Open support section | 6.56s | 23.6ms-43.6ms | ~99% faster |
| Comment action in support diff | 23.65s | 53.1ms | ~99.8% faster |
| Typing dispatch per character | 160ms-225ms | 1.4ms-4.4ms | ~97-99% faster |
| Mounted walkthrough diff views | 27-54 | 1 active step view | much less diff DOM work |

The after timings came from the profiled branch while testing the same large walkthrough. The final cleanup goes a little further structurally by rendering only the active step, but I have not rerun the full profiler trace after that last cleanup.

Validation:
- `vp check --fix`
- `vp test`
- `vp build`

`vp build` still reports the existing Vite chunk-size warning.


cpojer:
Thanks for the PR and for investigating the issue. #67 should fix most of this, so let's check back in once that is merged and released.

ericchernuka:
Thanks @cpojer for entertaining it!

cpojer:
Yeah no of course, Codiff should be fast and beautiful, and the new walkthroughs we shipped yesterday weren't fast yet. Sorry for the regressions!

PR #77, opened 2026-06-12
Add configurable `reviewCommentsPrefix` setting

henryqdineen:
The copied review comments markdown hardcodes `# Address these Review Comments` as a prefix. This bakes in an assumption about how the output will be used — it reads as an instruction from a third-party reviewer. In at least one instance, a coding agent that received pasted review comments assumed the comments came from someone else and treated them differently than if the user had written the instructions directly.

This adds a `reviewCommentsPrefix` setting so users can control the prefix. It defaults to the existing heading for backwards compatibility. Set it to any markdown string, or `""` to omit the prefix entirely.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/config/defaults.json
+++ b/config/defaults.json
@@ -10,6 +10,7 @@
     "lastRepositoryPath": "",
     "openAIModel": "gpt-5.3-codex-spark",
     "piModel": "pi-default",
+    "reviewCommentsPrefix": "# Address these Review Comments",
     "showOutdated": false,
     "showWhitespace": false,
     "theme": "system",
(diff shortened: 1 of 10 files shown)


cpojer:
One thing that is important to me is preserving behavior for people with existing config files: if `reviewCommentsPrefix` is not present at all, Codiff should use the default. If it is explicitly present as an empty string (`""`) in the user config, Codiff should append neither a prefix nor the two separator newlines.

cpojer (approved):
Rebased it for you.

PR #79, opened 2026-06-13
Fix syntax highlighting for .mts files

### LINKED ISSUE #78: .mts files do not receive syntax highlighting in diffs
Codiff appears to skip syntax highlighting for `.mts` files even though equivalent `.ts` files highlight correctly.

## Repro

Using Codiff v1.3.0 on macOS:

1. Create a repo with `src/example.mts`
2. Commit this baseline:

```ts
export type GreetingOptions = {
  readonly name: string;
  readonly excited?: boolean;
};

export function buildGreeting(options: GreetingOptions): string {
  const suffix = options.excited ? "!" : ".";
  return `Hello, ${options.name}${suffix}`;
}

console.log(buildGreeting({ name: "Codiff", excited: true }));
```

3. Modify it to:

```ts
import { inspect } from "node:util";

export type GreetingOptions = {
  readonly name: string;
  readonly excited?: boolean;
  readonly audience?: "human" | "agent";
};

export function buildGreeting(options: GreetingOptions): string {
  const suffix = options.excited === true ? "!" : ".";
  const audience = options.audience ?? "human";
  return `Hello, ${options.name} (${audience})${suffix}`;
}

const example = {
  audience: "agent",
  excited: true,
  name: "Codiff",
} satisfies GreetingOptions;

console.log(inspect({ message: buildGreeting(example) }));
```

4. Open Codiff on the repo.

## Expected

The `.mts` diff should use TypeScript syntax highlighting, like `.ts`.

## Actual

The `.mts` diff renders with no syntax highlighting.

## Control Case

Repeating the same repro with `src/example.ts` instead of `src/example.mts` produces syntax highlighting as expected.

## Notes

The bundled highlighter appears to include TypeScript and aliases for `ts`, `cts`, and `mts`, so this may be happening in the file classification or diff rendering path rather than in the grammar bundle itself.

scamden:
## Summary

- Register Node TypeScript module extensions with the diff highlighter
- Cover both `.mts` and `.cts` so they resolve to TypeScript syntax highlighting
- Add a focused regression test for the extension mapping

Fixes #78.

## Test plan

- `pnpm test src/__tests__/code-view-options.test.ts`
- `env GIT_CONFIG_GLOBAL=/private/tmp/codiff-empty-gitconfig pnpm test`
- `pnpm exec vp lint`
- `pnpm exec vp build`
- Manually launched patched Codiff against the `.mts` repro

### DIFF
--- a/src/__tests__/code-view-options.test.ts
+++ b/src/__tests__/code-view-options.test.ts
@@ -1,0 +1,8 @@
+import { getFiletypeFromFileName } from '@pierre/diffs';
+import { expect, test } from 'vite-plus/test';
+import '../lib/code-view-options.ts';
+
+test('registers Node TypeScript module extensions for syntax highlighting', () => {
+  expect(getFiletypeFromFileName('src/example.cts')).toBe('typescript');
+  expect(getFiletypeFromFileName('src/example.mts')).toBe('typescript');
+});
--- a/src/lib/code-view-options.ts
+++ b/src/lib/code-view-options.ts
@@ -1,10 +1,12 @@
-import { registerCustomTheme, type FileOptions } from '@pierre/diffs';
+import { registerCustomTheme, setCustomExtension, type FileOptions } from '@pierre/diffs';
 import dunkelTheme from '../themes/dunkel.json' with { type: 'json' };
 import lichtTheme from '../themes/licht.json' with { type: 'json' };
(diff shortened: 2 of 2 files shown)


cpojer (approved):
Good boi.

scamden:
lol

PR #80, opened 2026-06-15
Fix unstable file tree line counts for hidden whitespace changes

gumadeiras:
This fixes cases where file tree added/removed counts changed after selecting or scrolling to a file.

The root issue was that initial working-tree patches did not apply the `showWhitespace` setting, while lazy-loaded file content did. That made counts change after a file loaded more detail. Working-tree patch reads now use the same whitespace setting from the start, and changing `showWhitespace` reloads the current review state.

Also included:
- ignored stale lazy-load responses after reloads
- fixed empty patch-only sections so whitespace-only hidden changes do not count as real diffs
- refreshed tree row decorations when only line counts change
- updated config docs for `showWhitespace`

Validation:
- `vp check --fix`
- focused tests: 104 passed
- `vp build`

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,9 @@ create the file with defaults and open it in your editor. The file supports JSON
 trailing commas, includes a JSON schema reference for editor completion, and is watched while Codiff
 is running so changes apply to open windows.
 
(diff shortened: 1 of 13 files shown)


cpojer (commented):
Thanks!

PR #81, opened 2026-06-15
Fix Codiff icon on macOS Tahoe

gumadeiras:
This updates the macOS app bundle to include the Icon Composer asset catalog for Codiff’s app icon.

Tahoe uses the new `Assets.car` / `CFBundleIconName` path for app icons. The previous fix updated the legacy `.icns` file, which worked on Sequoia but still rendered as the small old-style icon on Tahoe.

Changes:
- Add the compiled `Assets.car` generated from `electron/icons/Codiff.icon`
- Set `CFBundleIconName` to `Icon` in the macOS app bundle
- Keep the existing `.icns` fallback for older macOS versions

Verified:
- `vp check --fix`
- `vp build`
- Tested the patched app bundle in a macOS Tahoe UTM VM; the Dock alias icon renders correctly with the full-size Codiff icon.

image


cpojer (approved):
That's awesome! Thank you.

Now the question is how do we update this icon to something prettier 😅

gumadeiras:
what do you think? 😅
codiff-bubbles-v1

cpojer:
I don't think very highly of it, you know I can make icons with an llm too :P

My thinking was mostly based on the color of the icon not matching the app at all. Ideally the logo adjusts to the same background color as the app has to have a consistent design language. Kinda like this from a vibes perspective (only focus on the colors):
CleanShot 2026-06-17 at 14 18 04@2x


I don't think we need added/removed lines in the logo like that though.

cpojer:
Did you notice that it looks kinda blurry? Why is this so hard 😭


Screenshot 2026-06-17 at 17 17 51

gumadeiras:
> Did you notice that it looks kinda blurry? Why is this so hard 😭
> 
> Screenshot 2026-06-17 at 17 17 51

looks fine on sequoia; pretty sure it's the liquid glass effect on Tahoe that makes all icons a bit blurry 🥲
i'm skipping tahoe entirely and waiting for golden gate to upgrade

PR #82, opened 2026-06-15
Update walkthrough guidelines

danfishgold:
When I use the `/codiff` skill in Claude Code (using Opus 4.8) it has a couple of issues.

It doesn't understand what hunkIds are and it wastes time going through Codiff's source code. I added a couple of examples to `walkthrough-guide.md` and removed the `hunkIds` description from the schema because it focused on the phrase "repository digest" and tried to find references to it in the source code.

Also, after generating the walkthrough json it would open it from the skills directory, and then Codiff wouldn't know to associate it with the repo's path. I've added the path as the last argument to the CLI in the examples in `SKILL.md` files

### DIFF
--- a/bin/walkthrough-guide.md
+++ b/bin/walkthrough-guide.md
@@ -59,6 +59,24 @@ choose.
 - Do not invent bugs, risks, tests, or validation. Describe what the diff and conversation
   actually support.
 
+## hunkId format
+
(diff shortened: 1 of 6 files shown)


cpojer (approved):
Makes sense, thank you.

PR #83, opened 2026-06-15
Add icons to comment buttons

danfishgold:
When looking over a PR I wasn't sure what would happen once I clicked "Ask" or "Comment". I think adding an icon to each of them would make it a lot clearer

before
Screenshot 2026-06-15 at 20 52 30

after
Screenshot 2026-06-15 at 20 51 12

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -1992,6 +1992,7 @@
 }
 
 .review-comment-action {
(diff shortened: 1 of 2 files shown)


cpojer (approved):
👍

PR #84, opened 2026-06-15
Make the header full width and move copy+PR actions to header

danfishgold:
## Current issues

The action buttons are currently hovering at the bottom of the window:

Screenshot 2026-06-15 at 20 58 14

This has some issues with the walkthrough's footer (as you can see in the screenshot) but also, and continuing my notes from #83, the GH buttons are unclear: at first I wasn't sure what they do so I clicked the checkmark and accidentally approved a PR 😬

An unrelated issue for me is that the window title is almost always truncated and when I have multiple Codiff windows open it makes it hard to understand which is which:

Screenshot 2026-06-15 at 21 01 13

## Proposed header

Right now when the sidebar is collapsed the header takes up the full width anyway so I thought I'd make it more consistent, and now that it's always there (and wide enough) there's room to put the floating buttons:

Screenshot 2026-06-15 at 21 07 05

## Related suggestions

* I wanted to add an "Open in GitHub" button but wasn't sure how you'd feel about it
* these buttons are probably too verbose. there wasn't a tooltip component already in the repo and adding one felt like too much for an already presumptuous PR from a first time contributor 🙃
* personally instead of submitting comments one by one and then accepting/rejecting a PR without a message I would prefer to go with what github's web UI encourages, which is to create all the comments as "drafts" and then publish them all at once under a review decision

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -156,13 +156,13 @@
 .app-shell {
   display: grid;
   grid-template-columns: 292px 0 minmax(0, 1fr);
+  grid-template-rows: auto minmax(0, 1fr);
   height: 100%;
   overflow: hidden;
 }
 
 .app-shell.sidebar-collapsed {
   grid-template-columns: minmax(0, 1fr);
-  grid-template-rows: auto minmax(0, 1fr);
 }
 
 .app-shell.sidebar-collapsed .sidebar {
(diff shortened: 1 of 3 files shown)


cpojer:
I like it, and I don't like the current buttons on the right bottom, but I'm not sure how I feel about a persistent top bar in Codiff. I think it would be much more fitting to use a persistent bottom bar, similar to what we already use in walkthroughs. What do you think?

danfishgold:
I think top bar is more conventional but bottom bar works too! either option would be better than the floating buttons in my opinion.
in any case I probably won't have time to work on this until the end of the month so feel free to close this PR or continue it without me 🥲
hopefully I'll have more time soon though. I've started using the app for work stuff and it's really great!

cpojer:
Ok, closing for now. Might get to it, if not, please send a PR in the future.

cpojer:
The copy button is still there for now, but the next release will have these buttons at the top of the scrollview.

PR #91, opened 2026-06-24
chore: remove unused `ts-node` dependency

camc314:
This PR removes the unused `ts-node` dependency. It was added in b2ab0b2, but was unused, it appears to have no uses.

Since node 24+ supports type stripping, we can use this instead, and skip downloading unnecessary deps.


cpojer (approved):
👍 

Thanks for the cleanup.

PR #101, opened 2026-06-30
Fix packaged app crash: include core/lib CJS modules

henryqdineen:
Fixes #100

I had [Claude](https://claude.ai/code) fix this. The `ignore` list in `forge.config.cjs` blanket-excludes `/core`, but `electron/narrative-walkthrough.cjs` requires `../core/lib/narrative-walkthrough-diff.cjs` at runtime. This converts `ignore` from an array to a function that whitelists `core/lib/*.cjs` while still excluding the rest of `core/` (React/TS source compiled into `dist/` by Vite).

### DIFF
--- a/forge.config.cjs
+++ b/forge.config.cjs
@@ -107,28 +107,38 @@ module.exports = {
     },
     executableName: 'codiff',
     ...(iconPath ? { icon: iconPath } : {}),
-    ignore: [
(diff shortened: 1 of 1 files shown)


cpojer:
Too late, already fixed in 1.6.1.

But thank you!

PR #102, opened 2026-07-05
Add commit submit keyboard shortcut

xixixao:
Adds a global Cmd/Ctrl+Enter handler in CommitView so the current commit form can be submitted from the keyboard.


cpojer (approved):
Nice catch.

PR #103, opened 2026-07-06
Stream commit output into a inner terminal

xixixao:
Before this MR I observed git commit output just flashing on error and disappearing.

Git commit can take a long time with pre-commit hooks.

With this MR, we show the git commit output in realtime, streamed into a terminal inside codiff.


cpojer:
Not excited to bring on xterm as a dep to Codiff tbh. What about libghostty?

cpojer:
Also, can you add a video?

xixixao:
I did it, but I'm tending towards this being a bad idea. The libghostty support for web is poor, the adapter I'm using here is not well-maintained. The scrollbar overlays content, I tried to patch it but it still showed some artifact from the native scrollbar. Resetting is painful as can be seen in code.

But it might be good enough for what we need here (just nice output).

https://github.com/user-attachments/assets/8676b6c1-f461-4a85-824d-d397eda1452b

cpojer:
Can you fix up the design a bit and add squircles etc. so it looks like a proper Codiff app box etc.?

xixixao:
Do you want to do it? You might have a better sense of what you want it to look like. It looked good enough to me.

cpojer:
I trust that you can ask fable and give it my comment and tell it squircles + make sure to follow the same design rules as in the rest of the repo + rebase the merge conflicts.

I'll do a final pass after.

xixixao:
Better?

https://github.com/user-attachments/assets/d2bbb3d8-cde3-4875-8e39-589bc0ecfd76

cpojer:
Sick. Thanks.

PR #106, opened 2026-07-06
Resolve the glab binary like other CLI integrations

jaqbec1:
## Summary

- resolve the `glab` binary in `electron/git-state/merge-request.cjs` the same way the Claude, Codex, OpenCode, Pi, and cloudflared integrations do: PATH lookup via the shared `agent-shared.cjs` helpers, with fallbacks to `~/.local/bin/glab`, `/opt/homebrew/bin/glab`, and `/usr/local/bin/glab`
- support a `CODIFF_GLAB_PATH` override, including a clear error when it points at a non-executable file
- expand the glab-not-found error to explain the searched locations and the override, matching the phrasing of the Claude/OpenCode messages
- add tests for the override, the invalid-override error code, and the guidance in the not-found message

The packaged `codiff` wrapper launches the app via `open -n`, so the GUI process inherits launchd's minimal PATH (`/usr/bin:/bin:/usr/sbin:/sbin`) and the previous bare `spawn('glab', ...)` failed with ENOENT even when glab was installed via Homebrew and authenticated, causing MR reviews to fall back to local changes.

Fixes #104.

## Validation

- `vp check --fix`
- `vp test core/__tests__/gitlab.test.ts` (14 passed)
- `vp build`

### DIFF
--- a/core/__tests__/gitlab.test.ts
+++ b/core/__tests__/gitlab.test.ts
@@ -1,4 +1,7 @@
+import { chmod, mkdtemp, rm, writeFile } from 'node:fs/promises';
 import { createRequire } from 'node:module';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
 import { describe, expect, test } from 'vite-plus/test';
 
 const require = createRequire(import.meta.url);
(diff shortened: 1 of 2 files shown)


cpojer:
Thank you!

PR #107, opened 2026-07-06
Refresh instead of hard reload

xixixao:
Controversial perhaps, but hard reloading the whole app on any change is annoying, it loses all state (which view I'm on, etc). It feels like a hack that worked to some point, but a proper refresh that only changes what's needed is the way to go?

Code is mostly by Fable, I just checked it works.


cpojer:
This is great.

PR #109, opened 2026-07-06
Dedupe equal author and committer in commit header

xixixao:
Easy simplification (mainly for local commits)

Screenshot 2026-07-06 at 21 10 16


cpojer:
Thank you!

PR #114, opened 2026-07-07
Read result event from newer Claude Code stream-array output

RazorSiM:
## Problem

Codiff's Claude Code integration fails for everyone on Claude Code **v2.1+** (repro'd on `2.1.202`): walkthrough generation and every "Ask" fail with **"Claude Code did not return JSON."**

`runClaude` spawns `claude -p --output-format json …` and then reads `envelope.result` / `envelope.structured_output`. That assumes `--output-format json` returns a single result envelope. Newer CLIs instead return a **JSON array of stream events**:

```jsonc
[
  { "type": "system", ... },
  { "type": "assistant", ... },
  { "type": "result", "is_error": false, "result": "{…}", "structured_output": {…} }
]
```

`JSON.parse(stdout)` still succeeds (an array is valid JSON), so the code silently fell through: `envelope.result` and `envelope.structured_output` were both `undefined`, `runClaude` resolved an empty string, and the caller threw the "did not return JSON" error.

Notably `claude --help` still documents `json` as *"single result"*, so this looks like an upstream CLI regression — this change is defensive and backward-compatible either way.

## Fix

Add `extractResultEnvelope(parsed)`, which collapses an array of stream events to its terminal `result` event (`findLast(type === 'result')`) and returns older single-object output unchanged. Apply it at the parse site in `runClaude`.

- **Old single-object output** → returned as-is (identical behavior).
- **New array output** → the terminal `result` event (with `result` + `structured_output`) is used.
- **Array-with-error** → `is_error: true` on the result event still rejects with the correct message (including the not-logged-in path).
- **Resultless array** → falls back to `{}`, matching prior behavior; genuine process failures still hit the non-zero-exit path before parsing.

## Tests

- Unit tests for `extractResultEnvelope`: array shape, single-object passthrough, and the empty-envelope fallback.
- A headless integration test driving `runClaude` with a fake `claude` that emits the array shape.

`vp check` (format + lint + typecheck) and `vp test` pass.

### DIFF
--- a/electron/__tests__/claude.test.ts
+++ b/electron/__tests__/claude.test.ts
@@ -8,6 +8,7 @@ const require = createRequire(import.meta.url);
 const {
   CLAUDE_NOT_FOUND_CODE,
   DEFAULT_CLAUDE_MODEL,
+  extractResultEnvelope,
   getClaudeCommand,
   isClaudeModelAvailabilityError,
   isClaudeNotLoggedInError,
@@ -16,6 +17,7 @@ const {
 } = require('../claude.cjs') as {
   CLAUDE_NOT_FOUND_CODE: string;
   DEFAULT_CLAUDE_MODEL: string;
+  extractResultEnvelope: (parsed: unknown) => any;
   getClaudeCommand: () => string;
   isClaudeModelAvailabilityError: (value: string) => boolean;
   isClaudeNotLoggedInError: (value: string) => boolean;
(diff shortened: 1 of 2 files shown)


cpojer:
I tried this with claude `2.1.195` and `2.1.202` and could not reproduce it. It seems like something is wrong with your setup, and your llm steered you towards the wrong fix for it.

RazorSiM:
> I tried this with claude `2.1.195` and `2.1.202` and could not reproduce it. It seems like something is wrong with your setup, and your llm steered you towards the wrong fix for it.

Thanks for pushing back :D

I tested again on 2.1.202 and could reliably reproduce. Digging into my global setup, I can "fix" it by switching the `verbose` option to `false`. With verbose on, `--output-format json` returns the full stream-event array instead of a single result, which is what breaks Codiff.

Up to you if you still value the change. I see two options:

1. Update this PR's description/comments. `--output-format json` documents as "single result," which isn't actually true when `verbose` is `true` so, IMHO, the change is correct as a defensive fix.
2. Force it at the call site with `--settings '{"verbose":false}'` -> tested and working.

PR #116, opened 2026-07-09
Update diffs to latest 1.3.0-beta.9

amadeus:
Ports things over to use thew new custom CodeView header api and also take advantage of the new hydration APIs.

You'll see the history updates to 2 different versions, Fable noticed a bit of an issue that was worth resolving in a secondary update.

Pretty much all this code was vibed, and since i don't have a strong understanding of the internals, unclear if this is the correct way forward on this stuff.

If it sucks, blame Fable 🤣


core/SharedWalkthroughApp.tsx
amadeus, line 1617: tbh, not sure why it deleted all this lol, just lmk if it should stay and i'll get it to add it back

cpojer:
arigatou!

PR #123, opened 2026-07-10
Add walkthrough persistence, resumable sessions, and staleness UI

RobinNagpal:
Timeout & sessions:
- Raise the walkthrough generation timeout well past ten minutes so large diffs (and resumed regenerations) are not killed early.
- Generate walkthroughs in a persistent, resumable Claude session; a PR's regeneration resumes the same session and passes the previous walkthrough so the model updates it in place. Session id is stored per repo + source.

Persistence & staleness:
- Persist the last walkthrough to ~/.codiff/walkthroughs along with the commit and working-tree signature it was generated against.
- On open, show the saved walkthrough instead of regenerating (both on launch and when opening the walkthrough tab); only generate when none is saved.
- Show how stale the saved walkthrough is (latest / N commits behind / diverged / working tree changed) with a Sync/Regenerate button in the walkthrough panel.

UI:
- Add a refresh button to the PR/source bar.

### DIFF
--- a/core/App.css
+++ b/core/App.css
@@ -712,6 +712,16 @@ html[data-codiff-platform='darwin'] .collapsed-sidebar-bar {
   background: rgb(127 127 127 / 0.2);
 }
 
+.sidebar-refresh-button {
+  flex: none;
+  margin-left: 4px;
+}
+
+.sidebar-refresh-button:disabled {
+  cursor: default;
+  opacity: 0.4;
+}
+
 .sidebar-resizer {
   -webkit-app-region: no-drag;
   border-left: 1px solid var(--sidebar-border);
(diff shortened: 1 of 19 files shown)


cpojer:
Hey! Thanks for sending a PR. This PR is doing way too much, and there are are various bugs in the implementation. You can already `cmd+r`, so I don't think we need to add a new UI element. I'm also not a fan of increasing timeouts. The latest release had a ~5x perf improvement for larger walkthroughs (10k line changes) and they should take no longer than about 20 seconds. If they do, please run the Codiff evals in the repo through your coding agent and ask it to optimize them – it's possible that the Codex variant works better than Claude right now for example.

I re-implemented walkthrough caching and shipped it as part of https://github.com/nkzw-tech/codiff/commit/f046e1c8d8613b55ce32228520602350c9975812.

PR #126, opened 2026-07-13
Branch + uncommitted vs main

xixixao:
Adds an option to compare the current uncommitted changes and any changes on the branch against main.

Screenshot 2026-07-13 at 17 33 42


cpojer:
Thank you!

PR #127, opened 2026-07-15
fix: prevent walkthrough comment layout loop

liangmiQwQ:
The walkthrough can enter a continuous layout invalidation loop after a Codex reply is added to a review comment. The comment editor and reply independently report different heights through one shared tracker, repeatedly rerendering the virtualized diff and blocking scrolling.

This PR observes the aggregate review thread instead, so its diff item is invalidated only when the thread's actual height changes. It also adds regression coverage for a tall Codex reply.

🤖 Generated with Codex


liangmiQwQ:
https://github.com/user-attachments/assets/15a5f493-1bbe-447e-846b-69bab0481c86

liangmiQwQ:
@codex review

cpojer (approved):
🙇‍♂️ Thank you. Will release today.

PR #128, opened 2026-07-15
Add simple diff mode to the codiff skill

rhart-rpadg:
## Summary

Adds a **simple diff mode** to the `codiff` skill so users can open Codiff's plain diff viewer without generating a narrative walkthrough.

- Plain `$codiff` still generates a narrative walkthrough (unchanged).
- `$codiff diff` (new) opens the raw diff viewer.

## What changed

Applied consistently across all four agent skill copies — **claude, codex, opencode, pi**:

**`scripts/open-codiff.mjs`** — new `--diff` flag:
- Spawns `codiff` without `-w` / `--walkthrough-file`, so the app opens its plain diff viewer.
- Defaults to the session repo's staged + unstaged changes; forwards any explicit target (commit, `HEAD`, branch, PR/MR, range, or path) verbatim.
- Still passes `--agent` and the session id so a follow-up walkthrough can reuse the conversation.

**`SKILL.md`** — routing + docs:
- `description` frontmatter now advertises `$codiff diff` / plain diff view.
- New *simple diff mode* bullet in **Choose The Mode** (triggers on `$codiff diff`, "plain diff", "raw diff", "just the diff").
- New **Simple Diff Mode** section with example commands.

## Verification

- All four launchers pass `node --check`.
- `--diff` produces a plain `codiff` invocation (no `-w`, no walkthrough file); defaults to the session repo; forwards targets like `HEAD`.
- Existing walkthrough mode still requires `--file`.
- The real arg parser (`bin/arguments.js`) reports `walkthrough: false` for the generated invocation, confirming plain-diff behavior.

### DIFF
--- a/claude/skills/codiff/SKILL.md
+++ b/claude/skills/codiff/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: codiff
-description: Open Codiff for a narrative code walkthrough, a blocking plan handoff, or a shared walkthrough URL. Use when the user writes "$codiff", "/codiff", "$codiff plan", "$codiff share", "show me codiff", "open Codiff", or asks to review a change or edit a plan in Codiff.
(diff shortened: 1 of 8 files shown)


cpojer:
I don't think we need to add this. I just did `$codiff no walkthrough` and it worked.

PR #130, opened 2026-07-16
fix(useAppWalkthrough): use real app version in shared walkthrough snapshot

luojiyin1987:
## Summary

The shared walkthrough snapshot hardcoded `codiffVersion: 'dev'`, so every published/shareable walkthrough reported version `dev` and lost the real build version.

This change threads the real version from the main process (`app.getVersion()`) into the renderer and uses it for the snapshot.

## Changes

- `electron/main.cjs` — inject `app.getVersion()` into the `codiff:getLaunchOptions` IPC handler and the `windowLaunchOptions` map (matches the existing headless share path).
- `core/types.ts` — add `codiffVersion: string` to `CodiffLaunchOptions`.
- `core/lib/app-constants.ts` — `defaultLaunchOptions` gets a `''` default.
- `core/app/hooks/useAppWalkthrough.ts` — accept `codiffVersion` and use it instead of `'dev'`.
- `core/App.tsx` — pass `launchOptions.codiffVersion` through.

## Validation

- `vp check --fix` — no warnings, lint, or type errors.
- `vp build` — succeeds.

Note: both `nkzw-tech/codiff` and the fork have issues disabled, so no tracking issue was created.

### DIFF
--- a/core/App.tsx
+++ b/core/App.tsx
@@ -332,6 +332,7 @@ export default function App() {
     walkthroughSharing,
     walkthroughUnread,
   } = useAppWalkthrough({
+    codiffVersion: launchOptions.codiffVersion,
     preferencesRef,
     state,
     stateGenerationRef,
--- a/core/__tests__/App-render.test.tsx
+++ b/core/__tests__/App-render.test.tsx
@@ -167,6 +167,7 @@ const createCodiffMock = (overrides: Partial<Window['codiff']> = {}): Window['co
     name: 'Reviewer',
   })),
   getLaunchOptions: vi.fn(async () => ({
+    codiffVersion: '',
     repositoryPathProvided: true,
     walkthrough: false,
   })),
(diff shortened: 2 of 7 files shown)


cpojer:
Thanks. Walkthrough shares are not a public feature at this time. Why did you send this PR?

luojiyin1987:
Thanks for clarifying. I noticed the hardcoded `codiffVersion: 'dev'` while reviewing the walkthrough sharing code and assumed it was an unintended placeholder that could affect shared snapshot metadata.

I wasn’t aware that walkthrough sharing is currently a private, unreleased feature. I submitted the PR proactively based only on the code I saw; it wasn’t prompted by access to or use of the feature.

I also realized afterward that the main process already overwrites `codiffVersion` with `app.getVersion()` before upload, so this change is unnecessary. I’ll close the PR. Sorry for the noise.
 @cpojer

PR #136, opened 2026-07-22
Add descriptive Codiff window titles

kurtextrem:
Hi!

I often have multiple Codiffs windows open at once, e.g. to compare different implementations and so on. Currently they all show up as "Codiff" when right-clicking the codiff icon.

The `title` that currently exists in the code wasn't applied because the `index.html` contains `<title>Codiff</title>`, and Chromium (Electron) applies that after the window loads. Now the plan mode sets its own title, and non-plan mode does so too.

### DIFF
--- a/electron/__tests__/window-title.test.ts
+++ b/electron/__tests__/window-title.test.ts
@@ -1,0 +1,54 @@
+import { createRequire } from 'node:module';
+import { expect, test } from 'vite-plus/test';
+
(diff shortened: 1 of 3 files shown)


cpojer (approved):
Awesome, thanks for the contribution!

cpojer:
I'm updating the titles to match the web service.

PR #138, opened 2026-07-25
fix: accept review URLs copied from a review tab

AbdelrahmanHafez:
This PR was written primarily with Opus 5 xhigh.

Pasting a link like `https://github.com/org/repo/pull/1728/changes#r4821` doesn't open the PR, Codiff falls back to treating it as a branch name so the walkthrough never gets generated.

`parseReviewUrl` anchored both the GitHub and GitLab patterns at the end of the path with only an optional trailing slash, so any segment after the number broke the match. The `#r4821` fragment was never the problem, `new URL` already puts that in `.hash`.

This PR makes the parser tolerant and canonicalizes to `https://github.com/owner/repo/pull/N` or `https://host/path/-/merge_requests/N` before anything downstream sees the URL. Now accepted:

- trailing tab segments: `/changes`, `/files`, `/commits/<sha>`, `/diffs`, `/pipelines`
- query strings and anchors: `?diff=split`, `#note_991`
- `www.github.com`
- links pasted without a scheme: `github.com/owner/repo/pull/1728`
- markdown autolinks: `<https://...>`
- a `.git` suffix on the repo or project path
- legacy GitLab URLs without the `/-/` separator: `/group/project/merge_requests/23`

Nothing that used to be rejected became valid. `pull/0`, non-numeric numbers, `/pull/` paths on non-GitHub hosts and non-http(s) schemes still return `null`, and a scheme-less paste only counts as a link when it starts with a dotted, host-shaped authority, so relative paths and refs aren't mistaken for URLs.

The same strictness existed in a few other spots, each with its own copy of the regex:

- `electron/git-state/pull-request.cjs` and `electron/window-identity.cjs` now delegate to `parseReviewUrl` instead of re-implementing it
- `electron/main/command-line.cjs` and `bin/arguments.js` stored the raw argument as `source.url`, so a tolerant parser on its own wouldn't have fixed anything, the raw link still reached the Git layer. Both store the canonical URL now, including the `CODIFF_PULL_REQUEST_URL` path
- `bin/codiff-app` classifies the argument before Electron ever sees it, so its shell regex needed the same treatment. It also recognizes `codiff pr <url>` directly now instead of falling through to a `gh pr view` branch lookup

`service/api.ts` already tolerated trailing segments and only ever sees URLs Codiff wrote itself, so I left it alone.

Tests go through each entry point: the parser, both CLIs, the shell launcher and window identity. They're in a separate commit ahead of the fix.

🍌 🍌 🍌 😁

### DIFF
--- a/bin/arguments.js
+++ b/bin/arguments.js
@@ -256,7 +256,8 @@ const getReviewProviderMarker = (arg) =>
       ? 'gitlab'
       : null;
 
-const isPullRequestUrlArgument = (arg) => parseReviewUrl(arg) != null;
+// Store the canonical form so pasted review tabs, queries and anchors never reach the Git layer.
+const parsePullRequestUrlArgument = (arg) => parseReviewUrl(arg)?.url ?? null;
 
 export const resolvePullRequestUrl = (repositoryPath, number, provider) =>
   resolveReviewUrl(repositoryPath, number, provider);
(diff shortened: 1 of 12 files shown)


cpojer (commented):
Thank you!

PR #141, opened 2026-07-25
fix(keymap): match Alt shortcuts on macOS through the active keyboard layout

AbdelrahmanHafez:
I was trying to use the `Alt+z` shortcut to trigger word wrap on macOS and didn't work, this PR fixes it.

Changes made by Opus 5 and Fable 5 xhigh, with review loop with Sol 5.6 xhigh. The diff is 2.2k lines, ~1.7k of which are tests.

-----

`Alt+z` never fires on macOS because Option+Z types `Ω` and the matcher compares `event.key`. Every `Alt` binding has the same problem, since there's no Cmd or Ctrl to suppress the composition.

`Alt` combos now fall back to the physical key when the character was composed away. Key positions come from `native-keymap` (the module VS Code uses) in the Electron main process: it reports what every key types plain and with Shift held and fires an event when the input source changes, and the renderer keeps a synchronous cache fed over IPC. So `Alt+z` follows the key that actually types `z` on QWERTZ, AZERTY, and Dvorak, `Alt+Shift+?` the key whose Shift types `?`, and a shifted keystroke can also be written by its key, VS Code style: `Shift+/` and `Shift+?` are the same shortcut.

The fallback is narrow: `Alt` on macOS only, only when the character was composed away, and only when that character matches no binding anywhere, so a character the layout produces natively always beats the key under it. Property tests walk every nameable character through both Shift states on US, German, AZERTY, Dvorak, and Russian layouts: no keystroke may fire two actions, and every combo must land on a key that really types its character.

Tried and discarded:

- Hardcoded US position tables: fire the shortcut on whatever key the layout put at that position instead.
- `navigator.keyboard.getLayoutMap()`: unshifted characters only and no change event, so layout switches had to be inferred from focus and keypress heuristics.
- Canonicalizing `Shift+/` through a static US pair table when no layout is available: quietly changed what existing combos match on non-US keyboards.
- Compensating the macOS ISO Backquote/IntlBackslash swap: measured on an ISO MacBook, the uncompensated data is already correct, and VS Code ships no compensation either.

With no layout available (the web build, and the Windows package until it builds on a Windows runner instead of cross-packaging on Linux), matching is exactly what shipped before this branch; failed native reads are rejected rather than applied, so they can't erase a good layout. Nothing changes on a US layout.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,11 +173,14 @@ Shortcuts match on the character your layout produces. <kbd>Option</kbd> is the 
 replaces the character it would otherwise produce (<kbd>Option</kbd>+<kbd>z</kbd> types `Ω`), so
 `Alt` shortcuts fall back to the key your layout puts that character on. `Alt+z` follows the key
 that types <kbd>z</kbd> on QWERTZ, AZERTY, and Dvorak rather than the one US keyboards label
-<kbd>z</kbd>. On a layout that types no Latin at all, such as Cyrillic or Greek, letters and digits
-stay at their US positions so those shortcuts remain reachable. Shifted `Alt` shortcuts such as
-`Alt+Shift+?` also use US positions, because the layout only reports the character a key produces
-unmodified. This all applies only to a keypress that matches no shortcut by character, so it never
-overrides one that does.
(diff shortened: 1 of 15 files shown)


cpojer (approved):
Cool!

PR #142, opened 2026-07-25
feat(cli): add shell completions for bash, fish and zsh

AbdelrahmanHafez:
Adds completions to codiff, changes made by Opus 5 xhigh with review iterations from GPT 5.6 Sol xhigh.

---

`codiff --completions <bash|fish|zsh>` prints a completion script for that shell, so completions come from the CLI itself instead of a checked-in copy per shell that goes stale.

```bash
# bash, in ~/.bashrc
source <(codiff --completions bash)

# fish, in ~/.config/fish/completions/codiff.fish
codiff --completions fish | source

# zsh, on the fpath
codiff --completions zsh > ~/.zsh/completions/_codiff
```

The scripts are generated from `flagDefinitions`, so a new flag is covered by adding it there. What each flag's value completes to comes from the placeholder in its help text: `<codex|claude|opencode|pi>` completes those choices, `<ref>` completes git refs, `<file>` completes paths, and `<id>` completes nothing. Tests pin that mapping so renaming a placeholder fails instead of silently dropping completion. Hidden flags stay out.

Arguments follow the documented grammar, `codiff [options] [<ref> | <pr> | <url>] [path]`. The first argument completes refs, and everything after it completes paths. `--branch`, `--commit` and `--plan` already name the source, so they carry `consumesSource: true` and the argument after them completes paths too.

Two things worth calling out:

`bin/codiff-app` is the helper the cask installs on the PATH, and it parses flags itself, so it delegates `--completions` to the Node entry the same way it already does for `--share`. Without that, `codiff --completions fish` would read `fish` as a ref and open a window.

Ref names are attacker-controlled in a cloned repository, and `git check-ref-format` accepts `$(touch${IFS}pwned)` as a branch name. `compgen -W` expands the words it is given, so passing refs to it runs them on TAB. The bash script matches refs as data and hands them back quoted with `printf %q`. fish and zsh don't have this problem, they treat command substitution output as data. There's a test that completes in a repository with such a branch and checks nothing ran.

Tests cover the generated scripts, the CLI flag and the `codiff-app` delegation, plus bash completion behaviour through `COMP_WORDS` and fish through `complete -C`. The fish and zsh tests skip when the shell isn't installed, so they run locally and stay quiet on CI. They also start their shells with `--noprofile --norc`, because `vp run` injects a library that makes bash read `~/.bashrc` even for `-c`, and anything that file prints ends up in the captured candidates.


cpojer (commented):
Sweeet!

PR #145, opened 2026-07-28
fix: open external links in the default browser

AbdelrahmanHafez:
Clicking the PR link in the top bar opened GitHub in an in-app Electron window. That window has no cookies or login session, so private repos render GitHub's 404 page:

image

The root cause is that no window open handler was registered, so Electron's default behavior (spawn a child window) applied to every `target="_blank"` link: the PR link, CI check links, and links in rendered markdown.

This PR adds a `setWindowOpenHandler` that denies in-app windows and hands http(s) urls to `shell.openExternal`, so links open in the user's default browser on macOS, Windows, and Linux. A `will-navigate` guard keeps the app window itself from being steered off the app, other schemes like `file:` and `javascript:` are dropped entirely, and rejections from `shell.openExternal` are handled so a failed launch can't crash the main process.

Changes made using Fable 5 xhigh, with review loop by Sol 5.6 xhigh.
I built the app with these changes and can confirm they work as intended.

### DIFF
--- a/electron/__tests__/external-links.test.ts
+++ b/electron/__tests__/external-links.test.ts
@@ -1,0 +1,120 @@
+import { createRequire } from 'node:module';
+import { expect, test, vi } from 'vite-plus/test';
+
+const require = createRequire(import.meta.url);
+const { attachExternalLinkHandling } = require('../external-links.cjs') as {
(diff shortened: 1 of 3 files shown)


cpojer:
Makes sense, thank you!

PR #147, opened 2026-07-28
feat: make switching repositories discoverable and show complete paths

AbdelrahmanHafez:
Based on feat/open-review-source-ui (#146), so the diff here includes those commits until it merges.

There was no way to switch repositories from within the app, only through the native menu bar. The repository label in the top bar also linked to the PR page, which the PR badge on the right already does. And paths rendered as fish-style abbreviations (`~/d/w/blog` instead of `~/dev/websites/blog`), which reads like line noise if you don't live in fish.

This PR:

- adds an Open Folder entry to the + menu and the command palette
- makes the repository label plain text with the full path in its tooltip, instead of duplicating the PR link
- shows complete repository paths everywhere, abbreviating only the home directory to `~`

The top bar label keeps the repository name visible when space runs out: the head of the path shrinks with an ellipsis in the middle, the last segment never truncates. Paths outside the home directory stay absolute, and Windows paths keep their drive prefix.

image

image

Created with Fable 5 xhigh, review iteration with Sol 5.6 xhigh. Built and tested and everything works as expected.

### DIFF
--- a/core/App.css
+++ b/core/App.css
@@ -3844,6 +3844,7 @@ diffs-container .review-comment-thread {
 
 .empty-panel code {
   color: var(--text);
+  overflow-wrap: anywhere;
 }
 
 .empty-panel-menu-path {
--- a/core/App.css
+++ b/core/App.css
@@ -768,17 +768,10 @@ html[data-codiff-platform='darwin'] .workspace-top-bar {
   white-space: nowrap;
 }
 
-button.review-top-bar-repository {
+.review-top-bar-repository-path {
   align-items: center;
-  background: transparent;
-  border: 0;
-  color: inherit;
-  cursor: pointer;
   display: flex;
   flex: 0 1 auto;
-  font: inherit;
-  padding: 0;
-  text-align: left;
 }
 
 .review-top-bar-repository-head {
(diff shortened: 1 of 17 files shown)


cpojer:
Thanks! Please rebase on main. I don't believe this should changed SharedWalkthroughApp (the web version) though? Because this would show local repo paths there in the wrong way. Please scope the changes only to the desktop app. Also I think extremely long repo names are clipped without an ellipsis as fixed-size tails cannot shrink. Maybe that could be fixed, too.

AbdelrahmanHafez:
Rebased on main. On the shared viewer, it was already showing local paths before this PR, just fish-abbreviated (`~/d/codiff-web` instead of `~/dev/codiff-web`), this PR only makes them complete. I kept the change there so the web viewer and the desktop app stay consistent, but if you feel strongly about it, I'm happy to scope it back to the desktop.

Long repo names are fixed too. The tail span was `flex: none` so it could never shrink and just got clipped by the parent, now it's capped at the slot width and gets its own ellipsis once the head has fully collapsed. A normal repo name still never truncates while there's a head left to shrink.

image
image
image

cpojer:
Thank you

PR #148, opened 2026-07-29
feat: add update notifications and one-click updates

AbdelrahmanHafez:
This PR adds update notifications and in-place updates, using GH releases as the source of truth.

This PR:

- Checks GitHub Releases once a day and caches the result in `~/.codiff/update-state.json`, shared by the app and the CLI
- Shows a small pill at the bottom of the sidebar when an update is available. One click applies it: macOS and Squirrel Windows download the update and restart into the new version, Linux and non-Squirrel Windows download the matching release asset and open it, then the pill asks you to quit to finish
- The pill tracks progress (updating, quit to finish, failed with one-click retry), and its tooltip says what clicking will do for the platform's update strategy
- Adds `codiff update` to the CLI, which upgrades through Homebrew when the cask owns the install, opens the app to update itself otherwise, and points at the release page for source checkouts
- Prints a one-line notice after `codiff --version` when a newer version is cached, implemented in the Node CLI and mirrored in the POSIX sh wrapper with the same version comparison and state validation so a broken or stale state file stays silent

<details>
<summary>Screenshots</summary>
image
image
</details>

<details>
<summary>Previous iterations</summary>
This iteration was a two-click update, which after some thought it felt like it can be simplified to match modern tools like VSCode and Codex, an update is just a single-click and self-explanatory. So I decided to match it.
image
image

</details>

Used Fable 5 xhigh with 20 iterations of review with Sol 5.6 xhigh.


cpojer:
This is awesome! Please make sure the button uses the same style as the <Button> component, for example `corner-shape: squircle`. We don't need to invent a new button design here.

Here are a few issues identified in review:

> 1. Windows updates don’t actually install. Windows releases deliberately contain only a ZIP and don’t include Squirrel. The fallback [downloads and opens the ZIP](https://github.com/nkzw-tech/codiff/blob/14ee3588f4bed6c49b70c8de50bb3aefef2b9778/electron/updater.cjs#L262-L288)⁠￼, then tells the user to quit to finish. Opening a ZIP does not replace the installed application.
> 2. The GUI bypasses Homebrew ownership on macOS. The CLI correctly detects the cask and runs brew upgrade, but the application always selects Electron’s updater on macOS. Since the Codiff cask does not declare itself self-updating, this can modify the Homebrew-managed app while leaving Homebrew’s installed-version receipt stale.
> 3. Downloaded installers are not integrity-checked. GitHub supplies SHA-256 digests for the current release assets, but the PR discards them, buffers the entire 119–175 MB asset in memory, writes it synchronously, and immediately opens it. An updater should verify the digest before opening an installer.
> 4. Linux asset matching ignores architecture. It selects the first .deb or .rpm, even though the updater receives process.arch. That will select the wrong installer once multiple Linux architectures are published.
> 5. Desktop-only CSS is added to shared App.css. The web app imports that stylesheet. Following the recent desktop/web split, the update-pill rules belong in Desktop.css.

For #2 we probably need to just update the homebrew tap to note that it is self updating etc.

cpojer:
Codex said this about auto-updating with Homebrew and on mac OS:

Add to the tap:

```ruby
livecheck do
  url :url
  strategy :github_latest
end

auto_updates true

depends_on arch: :arm64
depends_on :macos
```

> Homebrew specifically recommends auto_updates true when an app’s “Check for Updates” functionality downloads and installs the update, rather than merely opening a webpage. Homebrew can then compare the installed app bundle’s version against the cask version and avoid accidentally downgrading a self-updated app. [Homebrew Cask Cookbook](https://docs.brew.sh/Cask-Cookbook)⁠￼, [Homebrew FAQ](https://docs.brew.sh/FAQ)⁠￼
> 
> For #148, I would also make these corresponding changes:
> 
> * Make codiff update open the application’s self-updater on macOS, even when installed through Homebrew. Its current brew upgrade --cask codiff special case becomes unnecessary and may lag behind the newest GitHub release if the tap has not been updated yet.
> * Continue updating the cask’s version, sha256, and URL for fresh installations. auto_updates true does not eliminate normal cask maintenance.
> * Keep publishing only signed and notarized macOS ZIPs. Electron requires a signed application for Squirrel.Mac updates. [Electron autoUpdater documentation](https://www.electronjs.org/docs/latest/api/auto-updater/)⁠￼
> * Test an actual upgrade starting from brew install --cask nkzw-tech/tap/codiff, then verify:
>     * Codiff updates in place.
>     * The terminal-helper symlink still works.
>     * CFBundleShortVersionString reflects the new version.
>     * brew outdated and brew upgrade do not reinstall or downgrade it.
>     * brew uninstall --cask codiff still removes the updated app.
> 
> The existing build already has the important prerequisites: its bundle versions are correctly set to 1.9.2, and the local release build is Developer ID signed, notarized, and has a stapled ticket.
> 
> So I’d consider issue #2 resolved by adding auto_updates true, aligning codiff update with the in-app updater, and testing the Homebrew-installed upgrade path. It would not resolve #148’s Windows ZIP behavior, missing download verification, Linux architecture matching, or web CSS leakage.

AbdelrahmanHafez:
All addressed, and updated the button


image

-------

1. Windows now has its own `manual` strategy: since the releases deliberately ship only a ZIP, the update button opens the release page instead of downloading an archive that can't install itself. The status stays `available` because nothing was installed.
2. `codiff update` no longer special-cases brew, it hands every packaged install to the app's self-updater. On the tap side I'd add `auto_updates true` together with the first release that ships this updater, not before: `brew upgrade` skips `auto_updates` casks unless `--greedy`, so flipping it while 1.9.2 is current would leave existing brew installs with no update path at all.
3. Downloads now stream to disk through a SHA-256 hash and are verified against the digest GitHub publishes for the asset before anything is opened. A missing or mismatched digest deletes the file and surfaces an error. This also removes the full in-memory buffering and sync write.
4. Linux assets are matched against `process.arch` (amd64/arm64 for deb, x86_64/aarch64 for rpm), with a fallback to an unmarked asset. An installer marked for another architecture is refused rather than opened.
5. All the update pill CSS (and its variables) moved to `Desktop.css`.

The pill itself now renders through the shared `<Button>` component and only layers an accent fill on top, so it picks up the squircle corners and sizing from the shared styles.

cpojer:
Awesome! I think we are getting close with this PR. I have a bit more feedback from my agent's review here:

> Blocking: installer downloads can destroy an existing file
> 
> The updater downloads directly to the final filename in Downloads using createWriteStream(path). That truncates any existing same-named installer. If downloading or verification fails, the catch handler then deletes that path.
> 
> I reproduced this: an existing codiff_1.9.3_amd64.deb containing unrelated data was overwritten, failed verification, and disappeared entirely.
> 
> Relevant code: [electron/updater.cjs lines 326–345](https://github.com/nkzw-tech/codiff/blob/42addcf0203e23bce17d52814fdb866e667ad2dd/electron/updater.cjs#L326-L345)⁠￼.
> 
> It should download to a uniquely created temporary path and only expose/open that file after verification. It must never truncate or remove a pre-existing user file.
> 
> Smaller bug: disabling update checks can leave the notification visible
> 
> checkForUpdates: false prevents future scheduled checks, but the updater initializes its status from the cached state before considering that setting. Someone who disables updates after an update was cached will continue seeing the pill indefinitely.
> 
> Relevant paths:
> 
> * [Cached status initialization](https://github.com/nkzw-tech/codiff/blob/42addcf0203e23bce17d52814fdb866e667ad2dd/electron/updater.cjs#L137-L145)⁠￼
> * [Setting only guards scheduled checks](https://github.com/nkzw-tech/codiff/blob/42addcf0203e23bce17d52814fdb866e667ad2dd/electron/main.cjs#L1248-L1251)⁠￼
> 
> This is less severe, but it contradicts the setting’s description and centrally managed installations.

AbdelrahmanHafez:
Addressed both 👍 

----

1. The download now streams into a uniquely created staging directory inside the download folder and gets verified there. Only a verified installer moves into the folder itself, and it never overwrites anything: if the name is already taken it lands as `name (1).ext` instead, like a browser download would. A failed download or verification only ever deletes the staging copy, so a pre-existing `codiff_1.9.3_amd64.deb` with unrelated data survives both the write and the cleanup.
2. With `checkForUpdates: false` the updater no longer surfaces a cached update at launch, the pill starts idle regardless of what an earlier check persisted. An explicit "Check for Updates" from the menu still reports and shows the pill, since the setting only covers the automatic daily checks and their notifications.

The staging change also made us treat the asset name from the release metadata as untrusted, since it ends up as a path on disk. Names that aren't plain file names are now rejected before any filesystem operation. GitHub normalizes asset names on upload so this shouldn't ever trigger for a real release, the updater just no longer depends on it.

cpojer (approved):

cpojer:
Great work! Thanks for making this.

PR #149, opened 2026-07-29
fix: authenticate glab and agent CLIs through the login shell environment

AbdelrahmanHafez:
Follow-up to #144, which fixed `gh` authentication for GUI launches and left `glab` out of scope. Same story: a Dock-launched Codiff inherits launchd's minimal environment, the single-instance lock keeps it for the lifetime of the process, and `glab` never sees a `GITLAB_TOKEN` that only exists as a login shell export. GitLab requests fail with:

```
To get started with GitLab CLI, please run: glab auth login
```

While auditing the other spawn sites for the same gap, I found the agent CLIs have it too: `claude`, `codex` (exec and app-server), `opencode` (run and serve) and `pi` all spawn with plain `process.env`, so keys like `ANTHROPIC_API_KEY` or `OPENAI_API_KEY` that only live in the user's shell profile never reach them from a GUI launch.

This PR:

- Adds `getCommandEnvironment` to `login-shell-environment.cjs`: `{ ...(await getLoginShellEnvironment()), ...process.env }`, so the login shell only fills variables the process doesn't already have
- Spawns `glab` with it in `glabApi`, mirroring the `gh` fix
- Spawns the six agent CLI sites with it, keeping `OPENCODE_PERMISSION` as an explicit override on top for OpenCode
- Routes `runGhApi` through the same helper so the precedence rule is defined once
- Keeps the agent suites hermetic with a `beforeEach` that clears `SHELL`, since every runner now resolves the login shell; the new tests bring their own fake login shell

The commits are ordered test first, then fix, for both halves. The glab regression test fails on the exact error above before the implementation, and the agent tests assert the spawned environment through the existing command transport capture, including that the process environment wins over the login shell.

I looked at the remaining spawn sites and left them alone on purpose: `opencode export` reads local session files, `cloudflared` authenticates through a browser flow and an on-disk cert, and the git spawns work fine from GUI launches today.

Created with Fable 5 xhigh, reviewed by Sol 5.6 xhigh.

### DIFF
--- a/core/__tests__/gitlab.test.ts
+++ b/core/__tests__/gitlab.test.ts
@@ -118,6 +118,7 @@ process.stdin.on('end', () => {
   await using _environment = createTemporaryEnvironment({
     CODIFF_GLAB_PATH: fakeGlabPath,
     CODIFF_GLAB_TEST_CALLS: callsPath,
+    SHELL: undefined,
   });
 
   await callback(repo, async () =>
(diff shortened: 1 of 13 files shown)


cpojer:
Nice!

PR #150, opened 2026-08-09
Move the review mode tabs into the sidebar

timche:
# Move the review mode tabs into the sidebar

> First of three stacked title-bar PRs; the other two build on this one.

## Problem

- Walkthrough / Tree / History live in the title bar, but they only change what the sidebar shows.
- The title bar's center slot is the most prominent spot in the window, spent on a control whose effect is confined to the left column.

## What changed

- The tab list is now its own component, so the title bar and the sidebar can each render it.
- The desktop app renders it at the top of the sidebar, directly above the file filter; the title bar's center slot is freed for the next change in this stack.
- The shared walkthrough app keeps its tabs in the title bar — its mode set includes Comments and its sidebar is a different component, so moving it was out of scope here.
- `ReviewTopBar`'s mode props became a union: either all three are passed or none are, so a caller can't specify `modes` without `onModeChange`.

## Risks

- The sidebar is user-resizable. Labels collapse to icons below 200px via the existing `sidebar` container query; between 200px and the default width the three tabs share the row evenly.
- The `@media (max-width: 480px)` label-hiding rules are now scoped to `.review-top-bar`, so they only affect the shared app's title-bar tabs. A sidebar narrowed on a small window is handled by the container query instead.

## Omissions / not verified

- No change to tab semantics — `changeSidebarMode` and the walkthrough unread dot behave exactly as before. The existing tab-clicking test passes untouched, which is the evidence for that.

## Screen recording

https://github.com/user-attachments/assets/a92d22fe-039a-4b49-9ec5-f3ec1d13ba28

## Test plan

1. Open a repository with a walkthrough. **Expected:** Walkthrough / Tree / History appear at the top of the sidebar above the filter field, and no longer in the title bar.
2. Click each tab. **Expected:** the sidebar content switches exactly as before; Walkthrough still opens the narrative view in the main pane.
3. Trigger a walkthrough while on Tree. **Expected:** the unread dot still renders on the Walkthrough tab and clears when you switch to it.
4. Drag the sidebar divider narrower than ~200px. **Expected:** labels disappear, icons remain, all three tabs stay clickable and evenly spaced.
5. Drag it back to full width. **Expected:** labels return without the strip wrapping or clipping.
6. Collapse the sidebar entirely. **Expected:** no orphaned tab strip, no layout shift in the title bar.
7. Open a shared walkthrough in the browser. **Expected:** unchanged — its tabs, including Comments, are still in the title bar.

---

### DIFF
--- a/core/App.css
+++ b/core/App.css
@@ -721,6 +721,7 @@ html[data-codiff-platform='darwin'] .workspace-top-bar {
   grid-template-columns: minmax(0, 1fr) auto minmax(0, 1fr);
   min-width: 0;
   padding: 0 14px 0 78px;
+  position: relative;
 }
 
 .share-shell .review-top-bar {
@@ -740,11 +741,26 @@ html[data-codiff-platform='darwin'] .workspace-top-bar {
 }
 
 .review-top-bar-left {
+  grid-column: 1;
   justify-content: flex-start;
 }
 
 .review-top-bar-right {
+  grid-column: 3;
   justify-content: flex-end;
+}
+
+.review-top-bar-center {
+  align-items: center;
+  display: flex;
+  justify-content: center;
+  left: 50%;
+  max-width: min(38%, 320px);
+  min-width: 0;
+  pointer-events: none;
+  position: absolute;
+  top: 50%;
+  transform: translate(-50%, -50%);
 }
 
 .review-top-bar-repository-slot {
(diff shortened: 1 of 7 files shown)


cpojer:
Actually this used to be the design of Codiff until recently! The problem is that if we move the buttons in the sidebar, then the user expects only the sidebar will change – but it's changing the main view as well as the sidebar, so imho the top bar is the right spot for choosing the view.

Thanks for doing the exploration, this does look better than the previous version of Codiff 😅

PR #152, opened 2026-08-09
Show the copy review comments button in the title bar

timche:
# Show the copy review comments button in the title bar

> **Stacked on the centered-branch PR, which is itself stacked on the review-mode-tabs PR.** Until those merge this diff also shows their commits. Review it last; the diff narrows to just the change below once the parents land.

## Problem

- The copy-comments control was a floating circular button pinned to the bottom-right of the window.
- It overlapped diff content, sat far from everything else actionable, and gave no indication of how many comments it would copy.
- It also only existed once a comment did, so there was nothing to discover before you had written one and nothing to explain why it came and went.

## What changed

- Moved into the title bar's right-hand action area, sized like the other title-bar controls, showing the pending comment count.
- Always present, and disabled while the count is zero, so the control is discoverable before the first comment and never appears or disappears under the cursor.
- Reuses the sidebar toggle's own treatment — transparent at rest, muted text, grey wash on hover — so the two read as siblings. The disabled state follows the muted-plus-reduced-opacity convention used by the other disabled buttons. No new color tokens.
- Copying no longer changes the button's color — only the icon swaps to a checkmark.
- Tooltip now says what the button actually produces: *Copy review comments as markdown*.
- Spacing in the right-hand group matches the left group, so the gap between the source label and the copy button is the same 8px that separates the sidebar toggle from the open-source menu.
- The old floating action bar and its styles are gone; nothing else referenced them.

## Risks

- The count renders as `0` rather than being hidden, which keeps the button's width fixed so the title bar doesn't shift when the first comment lands.
- A permanently visible control is one more thing in the title bar; on narrow windows the source label and the button compete for the right group.
- While a review source is switching the button reads as empty and disabled rather than vanishing, so the layout holds still through the switch.

## Omissions / not verified

- The accessible label still carries the count (*Copy 3 review comments*), which is what a screen reader should hear; the visible tooltip describes the output format instead.

## Screen recording

https://github.com/user-attachments/assets/30d4b2ad-195b-42ee-90dc-b57d82410890

## Test plan

1. Open a repository with no comments. **Expected:** the button is in the title bar showing `0`, greyed out and not clickable.
2. Click it while disabled. **Expected:** nothing happens and the clipboard is untouched.
3. Add a review comment with a body. **Expected:** the button enables and reads `1`, without changing width or shifting the title bar.
4. Add two more. **Expected:** the count reads `3` and the tooltip reads *Copy review comments as markdown*.
5. Click it. **Expected:** the icon becomes a checkmark for ~2s, the color does **not** change, and the clipboard holds the rendered markdown.
6. Delete every comment. **Expected:** the button returns to `0` and disabled, still in place.
7. Create a comment and leave it empty. **Expected:** the button stays disabled — empty drafts don't count.
8. Switch review source while comments exist. **Expected:** the button reads `0` and disabled during the switch, then returns to the real count. No layout jump either way.
9. Hover and press the button next to the sidebar toggle. **Expected:** identical hover and active feedback on both.
10. Compare the gap either side of the title bar. **Expected:** the source label to copy button spacing matches the sidebar toggle to open-source menu spacing.
11. Toggle light and dark themes. **Expected:** enabled and disabled states both stay legible against the title bar.
12. Narrow the window until the source label and button crowd. **Expected:** neither overlaps the centered branch.

---

### DIFF
--- a/core/App.css
+++ b/core/App.css
@@ -3616,26 +3616,27 @@ diffs-container .review-comment-thread {
   word-break: break-word;
 }
 
+/* Matches .review-top-bar-icon-button, the sidebar toggle two controls over,
+   so the two read as siblings rather than as different kinds of control. */
 .copy-comments-button {
   -webkit-app-region: no-drag;
   align-items: center;
-  background: color-mix(in srgb, var(--sidebar-ref) 12%, var(--code-bg));
-  border: 1px solid color-mix(in srgb, var(--sidebar-ref) 38%, transparent);
-  border-radius: 14px;
-  color: var(--sidebar-ref);
+  background: transparent;
+  border: 0;
+  border-radius: 8px;
+  color: var(--muted);
   corner-shape: squircle;
   cursor: pointer;
   display: inline-flex;
   flex: none;
   font: 700 11px/1 var(--font-sans);
   gap: 5px;
-  height: 26px;
+  height: 28px;
   justify-content: center;
-  padding: 0 10px;
+  padding: 0 8px;
   transition:
-    background 150ms ease,
-    border-color 150ms ease,
-    transform 150ms cubic-bezier(0.34, 1.56, 0.64, 1);
+    background 120ms ease,
+    color 120ms ease;
 }
 
 .review-submit-button {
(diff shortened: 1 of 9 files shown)


cpojer:
Thank you.

PR #153, opened 2026-08-09
Add review comments with Mod+Enter instead of asking the agent

timche:
# Add review comments with Mod+Enter instead of asking the agent

> Independent of the title-bar stack — branches straight off `main` and shares no files with those PRs.

## Problem

- In a local review, `Mod+Enter` in a comment editor asked the agent. In a pull-request review the same chord submitted the comment.
- `Mod+Enter` is GitHub's "submit a comment" everywhere, so the local behaviour sends a comment to an LLM when the muscle memory says "post this".
- Asking the agent is the more expensive and less reversible of the two actions, which is the wrong way round for an accidental keypress.

## What changed

- `Mod+Enter` now always means *add this comment*. Pull-request mode submits as before; local mode finishes the comment the way clicking away does, so the draft is flushed and the editor closes.
- Asking the agent moved to a new configurable `askAgent` action, default `Mod+Alt+Enter`, and now works in pull-request mode too — previously the keyboard path existed only locally.
- The Ask button's tooltip shows the shortcut, and the shortcuts help panel lists the new action. `submitComment` is relabelled *Add comment*, since it no longer means "submit" in a local review.

## Risks

- **Behaviour change for existing muscle memory:** anyone who learned `Mod+Enter` to ask the agent now adds a comment instead. Both actions are rebindable in `codiff.jsonc` for anyone who wants the old mapping back.
- **Why not `Mod+Shift+Enter`:** GitHub already binds it in the Files changed tab, to *Submit a review comment*. Reusing it for "ask the AI" would recreate the same class of surprise this PR removes. `Mod+Alt+Enter` collides with nothing in the keymap, nothing GitHub binds, and nothing in the markdown editor (which claims `Mod+Alt+Digit0–6` for headings).
- The local add path blurs the focused element, because the editor handle exposes `focus` but no `blur`. Lexical turns that native blur into its own blur command, so the existing flush-and-cleanup path runs once, not twice.

## Omissions / not verified

- Saving an edit to an existing comment, and the general-comments composer, still use `Mod+Enter` and are untouched.
- The shared walkthrough and shared plan apps are untouched — neither has an ask-the-agent path.

## Screen recording

https://github.com/user-attachments/assets/95d08bcb-7700-4641-a61f-ecbb24fbcfaf

## Test plan

1. In a local review, start a comment, type a note, press `Mod+Enter`. **Expected:** the comment is kept and the editor closes. The agent is **not** asked.
2. Same, but press `Mod+Alt+Enter`. **Expected:** the agent is asked and focus stays in the editor.
3. Start a comment, type nothing, press `Mod+Enter`. **Expected:** the empty draft is discarded, exactly as clicking away does.
4. Press `Mod+Alt+Enter` on an empty draft. **Expected:** nothing happens — there is no note to ask about.
5. In a pull-request review, type a comment and press `Mod+Enter`. **Expected:** it submits, unchanged from before.
6. In a pull-request review, press `Mod+Alt+Enter`. **Expected:** the agent is asked — this is new.
7. Edit an existing comment and press `Mod+Enter`. **Expected:** the edit saves, unchanged.
8. Open the shortcuts help. **Expected:** *Add comment* on `Mod+Enter` and *Ask the agent* on `Mod+Alt+Enter`.
9. Rebind `askAgent` in `codiff.jsonc` and restart. **Expected:** the new binding applies and the Ask tooltip shows it.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,7 @@ counts; when it is `false`, Codiff hides those changes from the working-tree rev
     "prevSearchMatch": "Shift+Enter",
     "closeSearch": "Escape",
     "submitComment": "Mod+Enter",
+    "askAgent": "Mod+Alt+Enter",
     "discardComment": "Escape",
     "toggleSidebar": "Mod+Shift+b",
   },
--- a/config/defaults.json
+++ b/config/defaults.json
@@ -20,6 +20,7 @@
     "wordWrap": false
   },
   "keymap": {
+    "askAgent": "Mod+Alt+Enter",
     "closeSearch": "Escape",
     "commandBar": "Mod+Shift+p",
     "diffSearch": "Mod+f",
--- a/core/__tests__/ReviewCodeView-scroll.test.tsx
+++ b/core/__tests__/ReviewCodeView-scroll.test.tsx
@@ -1249,6 +1249,138 @@ test('review comment typing stays local until a comment action commits it', asyn
   expect(onAskCodex).toHaveBeenCalledWith('comment-1');
 });
 
+const renderLocalReviewComment = async ({
+  body,
+  onAskCodex,
+  onCommentDraftChange,
+  onDeleteComment,
+  onUpdateComment,
+}: {
+  body: string;
+  onAskCodex: () => void;
+  onCommentDraftChange?: () => void;
+  onDeleteComment?: () => void;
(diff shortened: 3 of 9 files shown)


cpojer:
Thank you.

PR #154, opened 2026-08-09
Add lightweight definition navigation

NikhilVerma:
## Summary

- add Command-click on macOS and Ctrl-click elsewhere to find likely local definitions
- underline navigable identifier-shaped tokens subtly while the modifier is held
- jump within Codiff when the exact definition line is present in the displayed diff; otherwise open the configured editor only when the search used the live working tree
- keep modifier affordances in sync with Pierre's rendered items through its public scroll and post-render lifecycle hooks
- add a deterministic temporary Git repository that doubles as a manual example and automated fixture

## Why a lightweight heuristic

A full Language Server Protocol integration would require per-language server/compiler installation, process lifecycle management, workspace indexing, configuration discovery, and distribution decisions. That is a large runtime and maintenance commitment for a local diff reviewer.

Tree-sitter is lighter than LSP, but it would still require bundling and updating grammars, maintaining language-specific declaration queries, and reconciling parsed trees with Codiff's working-tree, index, commit, range, branch, and pull-request snapshots.

This PR instead uses a bounded `git grep` over the relevant Git snapshot and applies conservative, language-aware declaration patterns. It has no daemon, index, compiler, grammar bundle, network call, or new runtime dependency. Search is limited to compatible file-extension groups, 20 matches per file, 1,000 matches and 256 KiB of output globally, a 1.5-second timeout, and 12 displayed candidates. A new search cancels the previous search for that window.

The scanner currently covers JavaScript/TypeScript, Python, Go, Rust, Java/Kotlin, C/C++/Objective-C, C#, Ruby, Swift, PHP, and shell scripts. The identifier affordance itself is language-neutral and skips common keywords plus syntax-highlighted comments and strings.

### Trade-offs

- Results are "likely definitions," not semantic guarantees. Overloads, re-exports, generated code, unusual declaration syntax, and dynamic languages can produce misses or multiple candidates.
- This does not provide references, type information, module resolution, or cross-repository navigation.
- Conservative patterns favor trustworthy candidates over broad recall. The patterns and ranking are isolated and covered by table-driven tests so they can be tuned incrementally.
- Editor fallback is restricted to live working-tree results. Historical results still jump when their exact line is present in the diff; otherwise Codiff marks them unavailable rather than opening a potentially different current-checkout file.

## Screenshots

Holding Command exposes subtle identifier affordances:

Command-held identifier highlights

Command-click shows likely definitions and marks this result as an in-diff jump:

Command-click definition result

## Reproduce manually

```bash
vpr example:definition-navigation
```

The temporary fixture demonstrates both paths:

- `formatGreeting` is defined in the diff and jumps inside Codiff.
- `DEFAULT_NAME` is defined in an unchanged file and opens in the configured editor.

## Validation

- `vp check --fix` — no formatter, lint, type, or warning failures
- `vp test` — 100 files and 990 tests passed
- `vpr build` — passed
- manually verified modifier highlighting across virtualized files, Command-click lookup, in-diff routing, and editor fallback

## AI assistance disclosure

OpenAI Codex was used for research, implementation, tests, documentation, screenshots, validation, and drafting this pull request description. The behavior and trade-offs were reviewed through interactive manual testing.

banana banana banana


cpojer:
Wow, sick. Here are a few comments from Codex review:

> 1. On isDefinitionNavigation in ReviewCodeView.tsx:
> 
> On macOS this also treats Control-click as definition navigation, but Control-click is the standard context-menu gesture. Could we use a shared platform-aware predicate—metaKey on macOS and ctrlKey elsewhere—for both clicking and modifier highlighting? Please add a regression test confirming macOS Control-click is not intercepted.
> 
> 2. On onOpenDefinition in App.tsx:
> 
> Definition search can inspect a commit, range, branch, or PR snapshot, but this fallback opens the file from the current checkout. If that snapshot is not checked out—or the file was changed, renamed, or deleted—the candidate’s line can point to different or nonexistent code. Could we restrict editor fallback to working-tree-compatible snapshots, or materialize/open the searched revision? In-diff navigation can still work for historical snapshots.
> 
> 3. On --max-count=${MAX_MATCHES_PER_FILE} in definition-search.cjs:
> 
> --max-count limits matches per file, so the total search is still effectively unbounded across a large repository. A common identifier can fill the shared 64MB Git buffer, after which every result is parsed and sorted. Repeated clicks also leave earlier Git processes running. Could we add a global output/result limit, timeout, and cancellation or coalescing? Searching the current file/directory before expanding repository-wide may also help.
> 
> 4. On the onFindDefinitions(...).then(...) call in ReviewCodeView.tsx:
> 
> This promise is not caught, and the request is not invalidated when the review source changes. An IPC rejection can become unhandled, while a slow result from the previous source can populate a stale popover after switching reviews. Please convert rejection into an unavailable result and close/invalidate the lookup when source changes or the component unmounts.
> 
> Optional packaging nit:
> 
> The current Forge packaging rules do not exclude examples/, so these screenshots and fixtures will ship in the desktop app. Could we exclude examples from packaging or keep the screenshots as GitHub-hosted documentation assets? The two PNGs alone add roughly 400KB.

NikhilVerma:
@cpojer Thanks — addressed all five points in 7cc8253.

- Definition navigation and modifier highlighting now share one platform-aware predicate. macOS uses Command only, so Control-click remains available for the native context menu; there is a regression test for that gesture.
- Editor fallback is now allowed only for live working-tree candidates. Historical/index results can still jump to an exact line in the displayed diff, but are disabled outside it instead of opening a potentially mismatched checkout.
- `git grep` is bounded globally to 1,000 records and 256 KiB, with a 1.5-second timeout. Starting another lookup cancels the previous process for that window, and window/render teardown cancels active work as well.
- Lookup rejections become an unavailable result, and pending lookups are invalidated on source changes and unmount.
- `examples/` is excluded from Forge packaging, with a config regression test.

I also updated the PR description to reflect the historical-snapshot restriction and the new search bounds.

Validation: `vp check --fix`, all 100 test files / 990 tests, and `vpr build` pass.

cpojer:
Thank you, this is awesome!

PR #155, opened 2026-08-10
Add button to copy code from collapsible markdown blocks

heckler1:
Agent-authored review comments put the prompt to paste into your agent inside a collapsed `<details>`, where the only way to get it out was to expand the section and hand-select the text. Code blocks render as read-only CodeMirror instances rather than `<pre>`, so there was nothing to copy from either.

The button copies only the fences belonging to its own level, since nested collapsibles render their own.

Written by Claude Opus 5, reviewed and validated locally by your friendly neighborhood human being.

Before: 
image

After:
image

### DIFF
--- a/core/App.css
+++ b/core/App.css
@@ -2180,11 +2180,30 @@ html[data-codiff-platform='darwin'] .sidebar {
 }
 
 .codiff-markdown-details > summary {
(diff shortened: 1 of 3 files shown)


cpojer:
Thank you.

PR #158, opened 2026-08-14
Copy an absolute file:line reference on Mod+Click

azhadsyed:
## TL;DR

Cmd+Click (macOS) or Ctrl+Click (Windows/Linux) on a diff line now copies `{absolute path}:{line number}` to the clipboard. This gives reviewers a pasteable location for an editor or a coding agent, while a plain click keeps creating a review comment.

**Risk:** Low
**Change type:** Additive

## Architecture

<details><summary>Why Meta/Ctrl, and how the repository root reaches the diff view</summary>

- **Modifier choice.** A plain click already creates a review comment, and Shift is claimed by the diff library for extending selections and expanding hunks. Meta and Ctrl were unclaimed. On macOS, Chromium remaps Ctrl+Click to a secondary click that fires `contextmenu` and never reaches the `click` handler, so the effective binding is Cmd+Click there and Ctrl+Click on Windows and Linux.
- **Precedence.** The modifier check sits in `onLineClick` after the existing interactive-element, collapsed-section, and text-selection guards, and it returns before comment creation. A modifier click never opens a comment editor, even when no repository root is available.
- **Prop threading.** The repository root reaches `ReviewCodeView` as a new optional `repositoryRoot` prop, fed from `RepositoryState.root` in the app's shared review props. The shared-walkthrough page never passes the prop, so Mod+Click stays inert there, since a reference relative to the author's checkout would mislead a web viewer.
- **Path joining.** The repository root carries the platform's separators while Git file paths always use forward slashes, so the copied reference joins with the root's own separator style.

</details>

## Testing

<details><summary>Unit tests drive the click handler directly, plus the full suite and build</summary>

Two new tests in `core/__tests__/ReviewCodeView-scroll.test.tsx` drive the `CodeView` `onLineClick` handler the same way the existing line-click test does:

- A meta-click with a `repositoryRoot` copies `{root}/{path}:{line}` to the clipboard and creates no comment.
- A ctrl-click without a `repositoryRoot` copies nothing and creates no comment.

`vp check --fix`, the full `vp test` run (99 files), and `vp run build` all pass locally. I did not manually exercise the gesture in the desktop app.

</details>

## AI Assistance

This PR was researched, implemented, and tested with Claude Code, per the contribution guidelines' disclosure policy.

banana banana banana

### DIFF
--- a/core/App.tsx
+++ b/core/App.tsx
@@ -1740,6 +1740,7 @@ export default function App() {
     onToggleCollapsed: toggleCollapsed,
     onToggleViewed: toggleViewed,
     onUpdateComment: updateComment,
+    repositoryRoot: state.root,
     searchQuery: diffSearchQuery,
     showWhitespace,
     source: state.source,
--- a/core/__tests__/ReviewCodeView-scroll.test.tsx
+++ b/core/__tests__/ReviewCodeView-scroll.test.tsx
@@ -1820,3 +1820,69 @@ test('line content clicks create review comments unless text is selected', async
   });
   expect(onCreateComment).toHaveBeenCalledTimes(3);
 });
+
+test('mod-clicking a line copies its absolute path and line number', async () => {
+  const writeText = vi.fn(async () => {});
+  Object.defineProperty(navigator, 'clipboard', {
+    configurable: true,
+    value: { writeText },
+  });
+  const onCreateComment = vi.fn();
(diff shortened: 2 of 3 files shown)


cpojer:
Hey! Thanks for sending a PR. Unfortunately we just merged a PR that makes this no longer works as cmd+click is now used for goto definition.

PR #159, opened 2026-08-18
fix: keep walkthrough scrolling stable across stop boundaries

bnjm:
Scrolling through a narrative walkthrough gets yanked backwards at every stop boundary, and clicking a stop in the table of contents can send the view scrolling somewhere else entirely — with the right walkthrough shape it degenerates into a scrolling loop the reviewer cannot escape.

## Root cause

Walkthrough stop-header items bake the current-stop selection into their CodeView item `version`:

```ts
version: getItemVersion(`${block.id}:walkthrough-header:${headerSelected ? 'selected' : 'idle'}`)
```

Crossing a stop boundary while scrolling changes `navigation.index`, which republishes two header items with new versions. CodeView treats a version change as a content change: it re-prepares the item, and the header's React-slotted annotation height falls back to the placeholder until the slot re-reports. The viewer's scroll anchoring then "corrects" the apparent shrink by scrolling backwards — by roughly the header's height, at every boundary, on every crossing. Driving the app over CDP shows the correlation exactly (each `STOP->n` transition below is one scroll event):

```
1699ms top=4681  STOP->2  REGRESSION from 4800
2801ms top=9361  STOP->3  REGRESSION from 9481
3889ms top=14042 STOP->4  REGRESSION from 14161
```

With realistic walkthroughs (multi-paragraph prose, one-hunk stops) the yank grows to ~230px per crossing; 120 wheel ticks that should land at 7200 only reached 6299. This is the same defect class 05e6cca fixed for file headers (selection styling must not feed item versions); the walkthrough header items kept the old pattern.

Two more defects compound the sidebar-click case:

- `syncSupportFromScroll` ignored the pending stop-scroll lock, so a click-initiated smooth scroll that crosses the support region flipped the mode to support mid-flight.
- The mode flip swapped the emitted block scroll target to the support branch, whose `request` comes from an independent counter — the shared dedupe ref in `ReviewCodeView` saw an unfamiliar number and issued a competing scroll command. Clicking a stop from the support region visibly reversed mid-flight and landed back in the support section.

## Fix

- Header item versions are stable (`${block.id}:walkthrough-header`). The current-stop accent is applied through the existing `codiff-selected-item` class in `onPostRender` plus a `.codiff-selected-item .wt-stop-block-header` rule, mirroring how 05e6cca handles file-header selection. `StopHeader`/`SupportHeader` lose their `current` prop; their content no longer depends on the active index at all.
- `useNarrativeNavigation` folds the stop and support scroll requests into one `scrollTarget: { index, kind, nonce }` with a single monotone nonce, and `getWalkthroughBlockScrollTarget` derives the target from that last explicit navigation action instead of the scroll-derived mode — a mode change caused by scrolling can no longer resurface an already-handled request as a new scroll command.
- `syncSupportFromScroll` respects the pending stop-scroll lock (released on arrival or by direct user scroll input, same as `syncIndexFromScroll`).

## Verification

Re-running the same CDP-driven experiments against the fixed build:

- 120 × 60px wheel ticks: **0 regressions**, final position **7200/7200**, active-stop tracking intact.
- The accent follows the active stop (exactly one accented header, always matching the arc) at every checked position.
- Clicking any stop — including from within the support region — lands at the correct block with zero scroll direction flips and settles in under a second.

Regression tests: walkthrough stop selection changes must not publish new header item versions (mirrors the existing file-selection test), `syncSupportFromScroll` honours the stop-scroll lock, stop/support requests share one nonce sequence, and `getWalkthroughBlockScrollTarget` tracks explicit navigation rather than scroll-derived mode.

`vp check --fix`, `vp test`, and `vp run build` pass (the pre-existing `web/`/`service/` lint findings on a clean tree are untouched).

---

This PR was investigated, written, and verified with Claude Code: the root-cause investigation drove the real Electron app over CDP with synthetic wheel/click input to measure the scroll regressions before and after the change, and the code, tests, and this description were AI-generated and human-reviewed.

banana banana banana

### DIFF
--- a/core/__tests__/ReviewCodeView-scroll.test.tsx
+++ b/core/__tests__/ReviewCodeView-scroll.test.tsx
@@ -119,6 +119,14 @@ const getCodeViewItemVersion = (id: string) =>
   (codeViewMock.lastItems.find((item) => item.id === id) as { version?: number } | undefined)
     ?.version;
 
+const getWalkthroughHeaderNode = (blockId: string) => {
+  const index = codeViewMock.lastItems.findIndex(
+    (item) => item.id === `${blockId}:walkthrough-header`,
+  );
+  expect(index).toBeGreaterThanOrEqual(0);
+  return codeViewMock.postRenderNodes[index];
+};
+
 const createLoadedMarkdownFile = (contents: string, fingerprint: string) => {
   const file = createChangedFileWithPatch(
     'plan.md',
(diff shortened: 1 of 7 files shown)


cpojer (approved):
Thank you!

bengolds:
I'll note that I'm still seeing this issue in v1.11.1!

cpojer:
Yeahhhh sometimes it still happens 😭

PR #163, opened 2026-08-21
Make file size limits configurable

brianangulo:
## Summary

The caps that decide whether a file loads automatically, loads on demand, or is skipped entirely are hard-coded in `electron/git-state/common.cjs`. Reviewing a change that touches a large file means editing the source and rebuilding, so this adds config settings for them.

| Setting | Default | Range | Effect |
| --- | --- | --- | --- |
| `eagerTextFileSizeMB` | 1 | 0.25–64 | Text file size that still loads automatically |
| `manualTextFileSizeMB` | 2 | 0.25–64 | Text file size that can still be loaded on demand; above this a file is never rendered |
| `imageFileSizeMB` | 32 | 1–512 | Image size rendered in a diff |
| `maxUntrackedFiles` | 1000 | 1–100000 | Untracked files listed individually before the rest collapse into one summary row |

Megabytes rather than bytes, because that is the unit someone raising a cap thinks in, and it keeps the schema's `minimum`/`maximum` readable. A `manualTextFileSizeMB` below `eagerTextFileSizeMB` would defer files Codiff already loads automatically, so the eager limit acts as its floor.

Defaults are unchanged, so nothing differs for anyone who does not set these.

## Implementation

`readConfig` reads and parses the config file on every call, so git-state reading config per file would cost a filesystem hit and a JSONC parse per file. Instead `common.cjs` keeps a module-level limits object seeded from `config/defaults.json` and exposes `getFileLimits()` and `setFileLimits(settings)`, converting megabytes to bytes once on the way in. Entry points push settings in once they have a config:

- `main.cjs` at `app.ready`, and again from `watchConfig`, so editing the config applies without a restart
- `headless-walkthrough-share.cjs` after each `readConfig()`

That keeps git-state free of filesystem reads and makes the limits testable without touching `~/.codiff`.

## Testing

`vp check --fix`, `vp test`, and `vp run build` pass (1010 tests passed, 4 skipped).

New tests cover normalization and fallbacks for non-numeric values, clamping at both ends of each range, the eager/manual floor, the megabyte-to-byte conversion in `setFileLimits`, and a 3 MB file that the default manual cap skips being deferred with `canLoad: true` once the cap is raised to 4 MB.

Also verified against a real `~/.codiff/codiff.jsonc` with the caps doubled that `readConfig` returns the raised values and `setFileLimits` converts them to the expected byte limits (2 MiB eager, 4 MiB manual, 64 MiB image).

## AI assistance

Written primarily with Claude Code (Claude Opus 5): research, implementation, tests, and this description.

banana banana banana


brianangulo:
Let me know if you have any questions @cpojer

cpojer:
Too much slop.

PR #165, opened 2026-08-30
Drop the port from remote-derived hosts so `glab --hostname` accepts them

oskarnyc:
# fix: drop port from remote host so `glab --hostname` accepts it

`parseRemoteUrl` used `URL.host` for `ssh://` / `https://` remotes, which includes the port
(`ssh://git@gitlab.example.com:2222/group/project.git` → `gitlab.example.com:2222`). That value is passed
to `glab api --hostname`, which rejects any `host:port` → `Error parsing --hostname: invalid hostname`
and `codiff mr <n>` fails with "Unable to read repository" on such GitLab instances.

Use `URL.hostname` instead. No behaviour change for remotes without a port (`host === hostname`);
the scp-style branch already had no port.

Test added: `parseRemoteUrl('ssh://git@gitlab.example.com:2222/group/project.git')` → `host: 'gitlab.example.com'`.

## Repro

Any repository whose `origin` remote is an `ssh://` (or `https://`) URL **with a port**, e.g.

```
git remote set-url origin ssh://git@gitlab.example.com:2222/group/project.git
codiff mr 1
```

Result (desktop window):

```
Unable to read repository
Error invoking remote method 'codiff:getRepositoryState':
Error: ERROR Error parsing --hostname: invalid hostname.
```

`glab` itself works fine in the same repo (`glab mr view 1` resolves the host from the remote).

### DIFF
--- a/electron/__tests__/review-source.test.ts
+++ b/electron/__tests__/review-source.test.ts
@@ -143,9 +143,9 @@ test('parseRemoteUrl drops a custom port from `ssh://` remotes', () => {
   });
 });
 
-test('parseRemoteUrl drops a custom port from `https://` remotes', () => {
+test('parseRemoteUrl keeps an explicit port on `https://` remotes', () => {
   expect(parseRemoteUrl('https://gitlab.example.com:8443/group/subgroup/project.git')).toEqual({
-    host: 'gitlab.example.com',
+    host: 'gitlab.example.com:8443',
     projectPath: 'group/subgroup/project',
     provider: 'gitlab',
   });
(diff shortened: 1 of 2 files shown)


cpojer:
Yo! Thanks for the contribution. This doesn't work as it strips the port both from ssh and https protocols, but if you strip them from https, it will break the identifier. Can you limit port stripping to ssh and update the https test?

oskarnyc:
Good catch, thanks — port stripping is now limited to `ssh://` remotes (transport port), while an explicit port on `https://` remotes is kept as part of the instance identifier. The https test now asserts the port is preserved. 11/11 passing.

cpojer:
Thanks!

PR #166, opened 2026-09-02
Add a commentOnLineClick setting to keep line clicks from opening comment drafts

azhadsyed:
## TL;DR
Adds `settings.commentOnLineClick` (default `true`). When set to `false`, clicking or dragging on diff lines no longer opens a review comment draft, and the gutter comment button remains the way to start one.

**Risk:** Low
**Change type:** Additive

## References
- Docs: [README § Configuration](https://github.com/nkzw-tech/codiff/blob/main/README.md#configuration) — the settings documentation this PR extends

## Background

<details>
<summary>Selecting text in a diff opened an empty comment draft under the line</summary>

To reproduce on main: open a diff, press the mouse inside a changed line, drag across a few characters, and release. An empty review comment draft opens under that line.

I select spans of code constantly while reading, so every selection left a stray draft to dismiss. The browser fires a click after a drag that starts and ends on one element, the diff library forwards it as a line click, and Codiff turns every line click into a draft. The text-selection guard only helps while the selection is still non-collapsed when that click fires.

With this change, a reader who prefers explicit commenting sets `commentOnLineClick` to `false` and keeps the gutter button (the comment control in the line-number gutter) as the one deliberate way to add a comment. Everyone else sees no change.

</details>

## Architecture

<details>
<summary>One setting gating two callbacks, with the gutter button left as the explicit path</summary>

- I gate the behavior behind a setting instead of detecting a text drag in the line-click handler. Whether the selection is still non-collapsed when the click arrives depends on timing, so a guard there is flaky. A setting is deterministic.
- The setting gates `onLineSelectionEnd` too. The code view starts a line-range selection only from a pointer-down in the line-number column (`startLineSelectionFromPointerDown` in `@pierre/diffs`) and reports it through that callback, so a number-column drag also ended in a draft. `onGutterUtilityClick` is untouched, so the gutter button stays a working, explicit path.
- With the setting off I also pass `enableLineSelection: false`, so a number-column drag no longer highlights a range that then does nothing.
- The default is `true`, so existing users see no change.

</details>

## Testing

<details>
<summary>Vitest cases for the gated callbacks and the config normalizer</summary>

- `core/__tests__/ReviewCodeView-scroll.test.tsx`: with `commentOnLineClick={false}`, a line click and a line-range selection end create no comment, the code view receives `enableLineSelection: false`, and a gutter button click still creates the comment at the expected line and section.
- `core/__tests__/config-defaults.test.ts`: the Electron reader keeps `false` when set and falls back to `true` on a missing or non-boolean value.
- `vp check`, `vp run build`, and the full `vp test` suite ran locally before pushing; CI is authoritative.
- By hand: set `"commentOnLineClick": false` in `~/.codiff/codiff.jsonc`, run this branch, then drag across text in a diff line and drag in the line-number column. No draft opens in either case, and the gutter button still opens one.

</details>

---

AI assistance: the code, tests, and this description were written with Claude Code (Claude Fable 5.1), directed and reviewed by the author.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,9 @@ is running so changes apply to open windows.
 Set `settings.showWhitespace` to `true` to show whitespace-only changes in diffs and file line
 counts; when it is `false`, Codiff hides those changes from the working-tree review state.
 
+Set `settings.commentOnLineClick` to `false` to stop clicks and drags on diff lines from opening a
+review comment draft. The comment button in the gutter still opens one.
+
 ```jsonc
 {
   "$schema": "https://raw.githubusercontent.com/nkzw-tech/codiff/main/core/config/codiff-config.schema.json",
@@ -141,6 +144,7 @@ counts; when it is `false`, Codiff hides those changes from the working-tree rev
     "claudeModel": "claude-sonnet-4-6",
     "codeFontFamily": "",
     "codeFontSize": 13,
+    "commentOnLineClick": true,
     "copyCommentsOnClose": false,
     "diffStyle": "split",
     "editorCommand": "",
--- a/config/defaults.json
+++ b/config/defaults.json
@@ -5,6 +5,7 @@
     "claudeModel": "claude-sonnet-4-6",
     "codeFontFamily": "",
     "codeFontSize": 13,
+    "commentOnLineClick": true,
     "copyCommentsOnClose": false,
     "diffStyle": "split",
     "editorCommand": "",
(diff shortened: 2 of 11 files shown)


azhadsyed:
Visual evidence: before this change, selecting text inside a diff line opened an empty review comment draft under the line. Screenshot of that state to be attached here.

cpojer:
Thank you for the PR. This is a regression. I noticed it as well when highlighting code and it is annoying. I will fix that without adding a new option.

PR #170, opened 2026-09-15
Keep local review comments when switching Markdown back to a diff

henryqdineen:
Clicking "View as Diff" on a Markdown file discarded every review comment in the session, including comments on unrelated files.

`toggleMarkdownPreview` calls `onRefreshMarkdown` when leaving preview mode for an editable working-tree section, so Markdown edits are flushed and the repository state is re-read. That refresh in `refreshMarkdownFile` then did `setReviewComments(getReviewCommentsFromState(orderedState))`. `getReviewCommentsFromState` only projects `state.reviewComments`, which holds backend comments and is populated for pull request sources. The refresh path only runs for `working-tree` and `branch-working-tree` sources, where that field is always empty, so the call replaced the entire comment list with `[]`.

Locally created comments never set `isReadOnly` and live only in React state, so all of them were lost: comment on a few lines of a JS file, toggle a Markdown file to its diff, and the comments are gone.

This adds `getRefreshedReviewComments`, which merges the refreshed backend comments over the local ones via the existing `mergeReviewComments` helper. Backend comments still win on matching ids, and local comments survive the refresh. The other `getReviewCommentsFromState` call sites are deliberate full reloads (initial load, source switch, whitespace preference change) and are unchanged.

### Testing

`core/__tests__/review-comments.test.ts` covers both directions: local comments (including empty drafts) survive a refresh that returns no backend comments, and read-only comments are replaced by the refreshed state. Both fail against the previous behavior. Also verified by hand in the app on a working tree with a JS and a Markdown file.


cpojer:
Thank you! New release coming in a bit.

PR #171, opened 2026-09-19
Fix walkthrough scroll snap-backs

matchai:
When a walkthrough step presents changes from the same file out of source order, scrolling can snap back at a hunk boundary. Codiff combined those hunks into one diff, but Pierre's scroll anchors assume source order within each diff.

Only coalesce forward runs of hunks. This preserves the authored walkthrough order while preventing the scroll jump.

**Before / After**

https://github.com/user-attachments/assets/8dcbfd27-cd98-45ca-b80d-47a78ebc07a5


cpojer:
Thank you!

